PR #62, opened 2019-01-16
Flutter github search example

Sach97:
## Status
**READY**

## Breaking Changes
NO

## Description
A few sentences describing the overall goals of the pull request's commits.

## Todos
- [x] Tests
- [x] Documentation
- [x] Examples

None


felangel:
Is this ready for review?

Sach97:
Yeah I think it is

felangel:
I’ll take a look shortly.

examples/flutter_search/lib/main.dart
```diff
@@ -0,0 +1,117 @@
...
+  // Create a text controller. We will use it to retrieve the current value
+  // of the TextField!
+  final _textController = TextEditingController();
+
+  final GithubSearchBloc _githubSearchBloc = GithubSearchBloc(
+      GithubService(GithubCache(), GithubClient(http.Client())));
+
+  @override
+  void initState() {
+    _githubSearchBloc.dispatch(TextChanged(text: ""));
```
felangel: why do we need this?

examples/flutter_search/lib/models/entities.dart
```diff
@@ -0,0 +1,76 @@
+import 'package:equatable/equatable.dart';
+
+class SearchResult extends Equatable {
+  final List<SearchResultItem> items;
+  final bool isPopulated;
```
felangel: why do we need `isPopulated` and `isEmpty`? Can't that all be derived from `items`?

examples/flutter_search/lib/models/entities.dart
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  @override
+  String toString() =>
+      'SearchResult { items: ${items.toString()}, isPopulated: $isPopulated, isEmpty: $isEmpty }';
+}
+
+class GithubUser extends Equatable {
+  final String login;
+  final String avatar_url;
```
felangel: the dart coding guidelines say to use camelCase for member variables so this should be `avatarUrl`

examples/flutter_search/lib/models/entities.dart
```diff
@@ -0,0 +1,76 @@
...
+      avatar_url: json['avatar_url'] as String,
+    );
+  }
+
+  @override
+  String toString() => 'GithubUser { login: $login, avatar_url: $avatar_url }';
+}
+
+class SearchResultItem extends Equatable {
+  final String full_name;
```
felangel: same comment here about using camelCase

examples/flutter_search/lib/main.dart
```diff
@@ -0,0 +1,117 @@
...
+            TextField(
+              controller: _textController,
+              onChanged: (text) {
+                _githubSearchBloc.dispatch(TextChanged(text: text));
+              },
+              decoration: InputDecoration(
+                  border: InputBorder.none,
+                  hintText: 'Please enter a search term'),
+            ),
+            _buildResult(state)
```
felangel: it's actually a bad pattern to write functions that return widgets because then Flutter doesn't know about the widget tree and can't perform optimizations. https://stackoverflow.com/questions/53234825/what-is-the-difference-between-functions-and-classes-to-create-widgets/53234826#53234826

As a result, I'd recommend moving the `_buildResult` logic directly into the build method.

examples/flutter_search/lib/github_service/github_service.dart
```diff
@@ -0,0 +1,28 @@
...
+
+class GithubService {
+  final GithubCache cache;
+  final GithubClient client;
+
+  GithubService(this.cache, this.client);
+
+  Future<SearchResult> search(String term) async {
+    if (cache.contains(term)) {
+      print("Result returned from cache");
```
felangel: can we remove the `print`?

examples/flutter_search/lib/bloc/simple_bloc_delegate.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:bloc/bloc.dart';
+
+class SimpleBlocDelegate extends BlocDelegate {
```
felangel: if this isn't being used we should remove it. Just in case it's not clear, you can use a `BlocDelegate` by doing something like:

```dart
void main() {
  BlocSupervisor().delegate = SimpleBlocDelegate();
  runApp(SearchApp());
}
```
and then you don't need to override `onTransition` in every bloc.

examples/flutter_search/lib/models/entities.dart
Sach97: What do you mean ? By doing this :
```dart
searchResultItems.length == 0 //isEmpty
```
I just copied an paste your angular example model but my suggestion would be to discard one of these two booleans because they reflect the same information.

felangel: yeah exactly my thinking. Since it's the same in the angular example I can update both separately.

Sach97:
Thank's for the review. I've made the requested changes :+1:

felangel:
can you please merge master into your branch first?

examples/flutter_github_search/android/app/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+android {
+    compileSdkVersion 27
+
+    lintOptions {
+        disable 'InvalidPackage'
+    }
+
+    defaultConfig {
+        // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
+        applicationId "com.example.fluttersearch"
```
felangel: shouldn't this be fluttergithubsearch?

examples/flutter_github_search/android/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,39 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.example.fluttersearch">
```
felangel: shouldn't this be fluttergithubsearch?

examples/flutter_github_search/android/app/src/main/java/com/example/fluttersearch/MainActivity.java
```diff
@@ -0,0 +1,13 @@
+package com.example.fluttergithubsearch;
```
felangel: shouldn't the directory be renamed to fluttergithubsearch?

PR #138, opened 2019-03-11
parametrised `stateChanged`

JJMinton:
## Status
**IN DEVELOPMENT** I wanted to submit this as a suggestion so I haven't thoroughly tested it yet.

## Breaking Changes
NO

## Description
I had a use case where the bloc state was an object that I wanted to update. By parametrising the condition that detects whether the state has changed (and hence rebuilds the widget tree), I am able to pass a function specific to the state object to detect these changes.

If there is reason why the state should be immutable, then by all means decline (hopefully with an explanation for my edification).

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_builder.dart
+++ b/packages/flutter_bloc/lib/src/bloc_builder.dart
@@ -17,11 +17,12 @@ typedef BlocWidgetBuilder<S> = Widget Function(BuildContext context, S state);
 class BlocBuilder<E, S> extends BlocBuilderBase<E, S> {
   final Bloc<E, S> bloc;
   final BlocWidgetBuilder<S> builder;
+  final bool(S, S) stateChanged;
 
-  const BlocBuilder({Key key, @required this.bloc, @required this.builder})
+  const BlocBuilder({Key key, @required this.bloc, @required this.builder, this.stateChanged})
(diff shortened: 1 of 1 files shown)


felangel:
@JJMinton thanks for bringing this up! The reason why bloc state is immutable is so that transitions can occur predictable (no state is mutated during/after a transition) and also to allow for efficient comparison of states. If the state were mutable then there would be no easy way to determine whether or not a state has changed and it would force the bloc to have to do a rebuild on every yield. Does that help?

JJMinton:
No worries, I thought there might have been something like that, and thanks for the quick reply.
Does that mean that if you wanted reasonably complex bloc state, the approach to take would be to create a class with a copy constructor?

felangel:
No problem! Yes, exactly 👍
Does that address your concerns?

JJMinton:
It does. It seems inefficient for larger objects but I suppose that's less important than state security.

Thanks for your time.

felangel:
My team and I have been using it in production and the largest state objects we’ve run into were mainly for form validation and even then we have not run into any performance issues. Creating a copy isn’t that expensive especially in comparison to a deep equality check but if you run into a scenario where it has an impact please let me know and we can work on how to resolve it! Thanks 👍

Issue #191, opened 2019-04-02
Can you add support for intelij idea 2019.1


MichelMelhem:

felangel:
Hey @michelmelhem 👋 

Does the current IntelliJ plugin not work? I’m not super familiar with IntelliJ plugins but my understanding was the plugin would be universal. Thanks!

MichelMelhem:
As you can see an this screenshot:
![image](https://user-images.githubusercontent.com/25546627/55476723-a12f9f00-5617-11e9-919e-79346a031ce1.png)

The last version of you plugin, the 1.1 is compactible just with intelij idea  2018.2.7 but since then a new version of intelij idea was relased the 2019.1.
So you need to recompile you app and specify that your plugin is compactible with intelij idea 2019.1

felangel:
Ah got it! We’ll try to have it updated ASAP. Thanks for bringing this up 👍

MichelMelhem:
Ok nice

jorgecoca:
I just submitted a new release, and now it is pending of approval by JetBrains:

<img width="890" alt="Screen Shot 2019-04-05 at 9 21 07 AM" src="https://user-images.githubusercontent.com/3237451/55634490-3cde1e00-5784-11e9-8dd7-becd9cb9c1e5.png">

jorgecoca:
New version has been released and approved here: https://plugins.jetbrains.com/plugin/12129-bloc-code-generator

Closing this issue. Thank you everyone!

PR #196, opened 2019-04-04
update intellij plugin to work with version 2019.1.0

felangel:
## Status
**READY**

## Breaking Changes
NO

## Description
- Update intellij version to `2019.1.0`
- Addresses #191 

None


extensions/intellij/intellij_generator_plugin/build.gradle
```diff
@@ -25,7 +25,7 @@ compileTestKotlin {
     kotlinOptions.jvmTarget = "1.8"
 }
 intellij {
-    version '2018.2.7'
+    version '2019.1.0'
```
jorgecoca: This should be `2019.1`. I'll update it and bump up the plugin version, and do a new release ;)

jorgecoca: Actually, this does not seem to be needed...

jorgecoca: Screen Shot 2019-04-05 at 7 36 20 AM

It seems it is defaulted to the latest version, so... I will remove it ;)

Issue #227, opened 2019-04-17
Bloc Plugin wont install in Intellij Ultimate 2019.1


cyrsis:
**Describe the bug**
Can not find plugin from the official Market

![image](https://user-images.githubusercontent.com/731087/56253703-1a3af600-60f0-11e9-80a1-3047d91e8e01.png)

Try manual install from Disk

![image](https://user-images.githubusercontent.com/731087/56253794-80277d80-60f0-11e9-901d-e0437839a86e.png)

Wish I can fix it and PR, but end up doesn't work too well

felangel:
@cyrsis thanks for reporting this! We’ll investigate and hopefully have a fix for this soon 👍

felangel:
@jorgecoca according to IntelliJ, the plugin should be compatible with IntelliJ Ultimate 2019.1. Any ideas what might be causing this?

cyrsis:
@felangel Works in Android Studio, Great plugin

felangel:
@cyrsis this should be addressed by #228

felangel:
Just pushed the update (v1.3) so it should be available in the next few days 👍 
Closing this for now but feel free to open another issue if the problem persists in v1.3

Issue #158, opened 2019-03-22
bloc Android plugin disabled


vendramini:
**Describe the bug**
Right click on any folder, New -> Bloc Generator -> New Bloc is always disabled.
```
Android Studio 3.3.2
Build #AI-182.5107.16.33.5314842, built on February 15, 2019
JRE: 1.8.0_152-release-1248-b01 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Linux 4.15.0-46-generic
```
Awesome work, guys =)

felangel:
@vendramini thanks for reporting this! 💯 

@jorgecoca have you been able to test the published plugin? I need to update my Android Studio still but I'll try to do that later this evening and see if I can reproduce this.

jorgecoca:
Nope, it is not disabled, see this:

<img width="1024" alt="Screen Shot 2019-03-22 at 8 59 27 PM" src="https://user-images.githubusercontent.com/3237451/54860254-85f29480-4ce5-11e9-9a7f-5b46d5a9d773.png">
<img width="647" alt="Screen Shot 2019-03-22 at 8 59 37 PM" src="https://user-images.githubusercontent.com/3237451/54860255-85f29480-4ce5-11e9-89b7-888cf41e31c4.png">
![bloc](https://user-images.githubusercontent.com/3237451/54860282-eaadef00-4ce5-11e9-9401-830c8195e703.gif)

felangel:
@vendramini can you give more details about how to reproduce the problem you're having?

vendramini:
I just downloaded the plugin directly from Android Studio.
Every folder with context menu shows me the "New Bloc" disabled.

How could I get more information for you?

felangel:
@vendramini I confirmed that it works for me as well
```
Android Studio 3.3.2
Build #AI-182.5107.16.33.5314842, built on February 15, 2019
JRE: 1.8.0_152-release-1248-b01 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
macOS 10.12.6
```

are you able to try on a different machine? The only difference I see is @jorgecoca and I are on macOS and you're on Linux 🤔

yy1300326388:
@felangel  I have the same problem on mac

![image](https://user-images.githubusercontent.com/8764899/54899059-8c615600-4f09-11e9-8cab-67946db0a775.png)


[...]

PR #228, opened 2019-04-17
Fix IntelliJ Plugin Compatibility Issues

felangel:
## Status
**READY**

## Breaking Changes
NO

## Description
- Explicitly specify the `idea-version` in the `plugin.xml` to address #227 
- Also addresses #158 by allowing users to generate a bloc regardless of the current directory.

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

None

### DIFF
--- a/extensions/intellij/intellij_generator_plugin/src/main/resources/META-INF/plugin.xml
+++ b/extensions/intellij/intellij_generator_plugin/src/main/resources/META-INF/plugin.xml
@@ -9,6 +9,13 @@
         ]]>
     </description>
 
+    <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
+    <idea-version since-build="145.0"/>
+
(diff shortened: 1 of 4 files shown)


extensions/intellij/intellij_generator_plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -9,6 +9,9 @@
         ]]>
     </description>
 
+    <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
+    <idea-version since-build="93"/>
```
jorgecoca: These IntelliJ docs are horrible...

felangel: I know haha 🤦‍♂️

felangel:
> Did you run this on an IntelliJ project? I think there are some generated files missing as part of that...

I ran and tested it on Android Studio and IntelliJ 👍

PR #534, opened 2019-09-23
Added 'const' for generated constructors

Mishanya840:
## Status
**READY**

## Breaking Changes
NO

## Description
Key word 'const' must be for linter prefer_const_constructors_in_immutables(https://dart-lang.github.io/linter/lints/prefer_const_constructors_in_immutables.html)

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Examples

## Steps to Test or Reproduce

```
- Run generate with Equatable flag
```

This PR will affect:

* Generate *State and *Event classes in mode with Equatable

### DIFF
--- a/extensions/intellij/intellij_generator_plugin/src/main/resources/templates/with_equatable/bloc_event.dart.template
+++ b/extensions/intellij/intellij_generator_plugin/src/main/resources/templates/with_equatable/bloc_event.dart.template
@@ -3,5 +3,5 @@ import 'package:meta/meta.dart';
 
 @immutable
 abstract class ${bloc_pascal_case}Event extends Equatable {
-  ${bloc_pascal_case}Event([List props = const <dynamic>[]]) : super(props);
+  const ${bloc_pascal_case}Event([List props = const <dynamic>[]]) : super(props);
 }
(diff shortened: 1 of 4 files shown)


anticafe:
If I remember correctly, from Dart 2.0, all definitions has `const` by default so it's safe to ignore it, for example: `Container(... child: Text(...))` instead `Container(... child: const Text(...))`.
So in your PR, how helpful `const` for constructor is?

Hacker-CB:
> If I remember correctly, from Dart 2.0, all definitions has `const` by default so it's safe to ignore it, for example: `Container(... child: Text(...))` instead `Container(... child: const Text(...))`.
> So in your PR, how helpful `const` for constructor is?

Linter option `prefer_const_constructors_in_immutables` warns about it.

felangel:
@Mishanya840 thanks for the contribution. I'm going to hold off on merging this until I have a chance to see whether or not this actually has a significant performance impact. I understand the linter warning but I want to be extra careful as this change would have a huge impact on the way developers use Equatable in their projects.

I will put together some performance tests later this week and will follow up with the results 👍

felangel:
This was incorporated into https://github.com/felangel/bloc/commit/3b1e32bc61a6015638d41525c957e91182aabea0, https://github.com/felangel/bloc/commit/895f4757bb83b76dab543b81bcfb8c2216ec897f, and https://github.com/felangel/bloc/commit/3708d3edfad5908eb59fc41e4de8cb8b9e244d13.

Closing and thanks for bringing this up! 🙏

Issue #558, opened 2019-10-09
[Proposal?] 'Stream<State> get state' is confusing


bernaferrari:
I've been developing a relatively large app using Bloc and I think my top 1 issue with Bloc is with state. No, not what you are thinking. I mean, the state variable/getter method.
![image](https://user-images.githubusercontent.com/351125/66448646-93c0fa80-ea28-11e9-8e9c-f65af5716f77.png)

It has happened to me once, then again, then again, then today. I always cast it, like `(currentState as ContrastLoadedState)` and I keep trying state before realising my mistake. Just check the screenshot above. It makes no sense for me that `currentState` returns a `State` while `state` returns a `Stream<State>`.

In BlocBuilder you get a state:
https://github.com/felangel/bloc/blob/805cf865fe0972e7f1bd24f08598d624066434fb/examples/flutter_shopping_cart/lib/catalog/my_catalog.dart#L52

In your own sample you use state = currentState:
https://github.com/felangel/bloc/blob/805cf865fe0972e7f1bd24f08598d624066434fb/examples/flutter_shopping_cart/lib/cart/bloc/cart_bloc.dart#L31

I think you could change state to stateStream and currentState to state or maybe just rename state to stateStream, that way there wouldn't be a state and IntelliJ/VSCode would be smart enough to just suggest currentState.

I'm not sure if this has been a source of bugs/confusion in the past, but it has been a great frustration for me.

felangel:
Hi @bernaferrari 👋 
Thanks for opening an issue!

I'm glad you brought this up and I agree the naming should be improved. What are your thoughts on renaming `state` to `states` and leaving `currentState` as is?

bernaferrari:
I don't know, I guess it would also be a good change? I personally would probably change it to `stateStream` or something like this, to keep in track with `_stateSubject` which already exists, but `stateStream` is not a perfect name, since it's not obvious what it should be doing from its name, so I guess `states` would also work fine.

felangel:
@bernaferrari I've discussed this with the team and we're currently split between:

- `bloc.states.listen`
- `bloc.stream.listen`
- `bloc.stateStream.listen`

What are your thoughts?

basketball-ico:
@felangel 
I prefer `bloc.states`
> Bloc: Takes a [Stream] of [Event]s as input and transforms them into a [Stream] of [State]s as output.

felangel:
[...]

PR #572, opened 2019-10-13
Refactor Bloc to extend Stream<State>

felangel:
## Status
**READY**

## Breaking Changes
YES

## Description
- Addresses #558 

### Refactor `Bloc` to extend `Stream<State>`
```dart
bloc.state.listen(print);
```

becomes

```dart
bloc.listen(print);
```

### Rename `currentState` to `state`

```dart
bloc.currentState
```

becomes

```dart
bloc.state
```

## Todos
- [X] Tests
- [ ] Documentation (will be added in a separate PR to avoid confusion before the update is published)
- [x] Examples

- Breaking change to the bloc API which will require a major version bump + updates to `flutter_bloc` and `angular_bloc`

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -6,24 +6,51 @@ import 'package:rxdart/rxdart.dart';
 
 /// Takes a [Stream] of [Event]s as input
 /// and transforms them into a [Stream] of [State]s as output.
-abstract class Bloc<Event, State> {
+abstract class Bloc<Event, State> extends Stream<State> {
   final PublishSubject<Event> _eventSubject = PublishSubject<Event>();
 
   BehaviorSubject<State> _stateSubject;
 
-  /// Returns [Stream] of [State]s.
-  /// Usually consumed by the presentation layer.
-  Stream<State> get state => _stateSubject.stream;
+  /// Returns the current [State] of the [Bloc].
+  State get state => _stateSubject.value;
 
   /// Returns the [State] before any [Event]s have been `dispatched`.
   State get initialState;
 
-  /// Returns the current [State] of the [Bloc].
-  State get currentState => _stateSubject.value;
+  /// Returns whether the `Stream<State>` is a broadcast stream.
+  bool get isBroadcast => _stateSubject.stream.isBroadcast;
 
   Bloc() {
     _stateSubject = BehaviorSubject<State>.seeded(initialState);
     _bindStateSubject();
+  }
+
+  /// Adds a subscription to the `Stream<State>`.
+  /// Returns a [StreamSubscription] which handles events from the `Stream<State>`
+  /// using the provided [onData], [onError] and [onDone] handlers.
+  StreamSubscription<State> listen(
+    void onData(State value), {
+    Function onError,
+    void onDone(),
+    bool cancelOnError,
+  }) {
+    return _stateSubject.stream.listen(
+      onData,
+      onError: onError,
+      onDone: onDone,
+      cancelOnError: cancelOnError,
+    );
+  }
+
+  /// Returns a multi-subscription stream that produces the same events as this.
+  Stream<State> asBroadcastStream({
+    void onListen(StreamSubscription<State> subscription),
+    void onCancel(StreamSubscription<State> subscription),
+  }) {
+    return _stateSubject.stream.asBroadcastStream(
+      onListen: onListen,
+      onCancel: onCancel,
+    );
   }
 
   /// Called whenever an [Event] is dispatched to the [Bloc].
@@ -136,9 +163,9 @@ abstract class Bloc<Event, State> {
       return mapEventToState(currentEvent).handleError(_handleError);
     })).forEach(
       (State nextState) {
-        if (currentState == nextState || _stateSubject.isClosed) return;
+        if (state == nextState || _stateSubject.isClosed) return;
         final transition = Transition(
-          currentState: currentState,
+          currentState: state,
           event: currentEvent,
           nextState: nextState,
         );
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -18,7 +18,7 @@ abstract class Bloc<Event, State> extends Stream<State> {
   State get initialState;
 
   /// Returns whether the `Stream<State>` is a broadcast stream.
-  bool get isBroadcast => _stateSubject.stream.isBroadcast;
+  bool get isBroadcast => _stateSubject.isBroadcast;
 
   Bloc() {
     _stateSubject = BehaviorSubject<State>.seeded(initialState);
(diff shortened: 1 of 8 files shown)


felangel:
@brianegan can you take a look? I'm planning to open a separate PR to implement `Sink<Event>` 👍

packages/bloc/lib/src/bloc.dart
```diff
@@ -6,26 +6,53 @@ import 'package:rxdart/rxdart.dart';
...
+  /// Returns the current [State] of the [Bloc].
+  State get state => _stateSubject.value;
 
   /// Returns the [State] before any [Event]s have been `dispatched`.
   State get initialState;
 
-  /// Returns the current [State] of the [Bloc].
-  State get currentState => _stateSubject.value;
+  /// Returns whether the `Stream<State>` is a broadcast stream.
+  bool get isBroadcast => _stateSubject.stream.isBroadcast;
```
brianegan: Can just use `_stateSubject.isBroadcast` as a shortcut. This will always return `true`, since Subjects are all broadcast streams

packages/bloc/lib/src/bloc.dart
```diff
@@ -6,26 +6,53 @@ import 'package:rxdart/rxdart.dart';
...
+    return _stateSubject.stream.listen(
+      onData,
+      onError: onError,
+      onDone: onDone,
+      cancelOnError: cancelOnError,
+    );
+  }
+
+  /// Returns a multi-subscription stream that produces the same events as this.
+  Stream<State> asBroadcastStream({
```
brianegan: No need for this method -- it's already a broadcast stream since it's based on the Subject's stream!

packages/bloc/lib/src/bloc.dart
```diff
@@ -6,26 +6,53 @@ import 'package:rxdart/rxdart.dart';
...
+  /// Adds a subscription to the `Stream<State>`.
+  /// Returns a [StreamSubscription] which handles events from the `Stream<State>`
+  /// using the provided [onData], [onError] and [onDone] handlers.
+  StreamSubscription<State> listen(
+    void onData(State value), {
+    Function onError,
+    void onDone(),
+    bool cancelOnError,
+  }) {
+    return _stateSubject.stream.listen(
```
brianegan: `_stateSubject.listen` also works as a shortcut.

packages/bloc/test/bloc_test.dart
```diff
@@ -83,13 +87,21 @@ void main() {
...
         simpleBloc.dispatch('event1');
         simpleBloc.dispatch('event2');
         simpleBloc.dispatch('event3');
       });
+
+      test('isBroadcast returns true', () {
+        expect(simpleBloc.isBroadcast, isTrue);
+      });
+
+      test('asBroadcastStream returns a broadcastStream', () {
```
brianegan: Could remove this test if ya remove `asBroadcastStream`, and ya can see here it's already a broadcast stream by default since it's using the subject under the hood which is itself a broadcast stream!

packages/bloc/test/bloc_test.dart
```diff
@@ -183,6 +197,14 @@ void main() {
         await Future<void>.delayed(Duration(milliseconds: 120));
         complexBloc.dispatch(ComplexEventC());
       });
+
+      test('isBroadcast returns true', () {
+        expect(complexBloc.isBroadcast, isTrue);
+      });
+
+      test('asBroadcastStream returns a broadcastStream', () {
```
brianegan: Can be removed

packages/bloc/test/bloc_test.dart
```diff
@@ -298,13 +320,21 @@ void main() {
...
         counterBloc.dispatch(CounterEvent.increment);
         counterBloc.dispatch(CounterEvent.increment);
         counterBloc.dispatch(CounterEvent.increment);
       });
+
+      test('isBroadcast returns true', () {
+        expect(counterBloc.isBroadcast, isTrue);
+      });
+
+      test('asBroadcastStream returns a broadcastStream', () {
```
brianegan: Can be removed

packages/bloc/test/bloc_test.dart
```diff
@@ -423,6 +455,14 @@ void main() {
 
         asyncBloc.dispatch(AsyncEvent());
       });
+
+      test('isBroadcast returns true', () {
+        expect(asyncBloc.isBroadcast, isTrue);
+      });
+
+      test('asBroadcastStream returns a broadcastStream', () {
```
brianegan: Can be removed

felangel:
Thanks for the feedback @brianegan. Can you please re-review? 😄

larssn:
Ouch, could you please notify of breaking changes before pushing big stuff like this? Thanks

larssn:
Maybe introduce a temporary `get currentState => state;`until people have had a chance to switch over. Our code has 1000+ errors since getting this new release. 😅

bigworld12:
@larssn well, it had a major version pump, so you shouldn't have gotten this via the usual pub get

tenhobi:
@larssn according to [semver](https://semver.org/#spec-item-4): v0.x.y should not be considered as stable. Keep it in mind for other packages you use as well.

Btw. first major release v1.0.0 is coming https://github.com/felangel/bloc/pull/572 !

larssn:
@bigworld12 
Going from 0.15.x to 0.16.x is a minor version bump.
Dart is using the standard `{major}.{minor}.{bugfix}`, see https://dart.dev/tools/pub/versioning

@tenhobi You're right, but semver is only respected by some package authors, and not others. I haven't seen mention that this package was not fit for production, or in alpha, for that matter.

In fact, there's a forum thread entitled "Are you using bloc in production". I know the BMW guys are using it in production, because they're stating so in that thread.

The API has changed a few times now. This time was a bit harder to swallow that the last one, as this conflicts with the lint rule `dart(close_sinks)`:

Example:
Doing stuff like `final authBloc = Provider.of<AuthenticationBloc>(context);` will now complain about authBloc being a stream that should be closed.

I hope I'm not coming off as angry, because that's not the case. I still consider the package essential for complex state management. 👍🏻

tenhobi:
@larssn In the link you provided, it is [mentioned](https://dart.dev/tools/pub/versioning#semantic-versions) this: 😃 

> Although semantic versioning doesn’t promise any compatibility between versions prior to 1.0.0, the Dart community convention is to treat those versions semantically as well. The interpretation of each number is just shifted down one slot: going from 0.1.2 to 0.2.0 indicates a breaking change, going to 0.1.3 indicates a new feature, and going to 0.1.2+1 indicates a change that doesn’t affect the public API.

If the package was updated from 0.15.x to 0.16.x (dunno if it was, that's what I see in your post) then it is concidered as a major change by Dart.

Check [caret syntax](https://dart.dev/tools/pub/dependencies#caret-syntax), it should work with `^` as this: `^0.15.2` is equivalent to `'>=0.15.2 <0.16.0'`.

larssn:
Lol, well, that certainly sounds different from standard semver then. 😅

Can't believe I missed that! So Dart is something like `{super_major}.{major}.{minor}+{bugfix}`.

felangel:
@larssn sorry for any inconvenience but as everyone else stated you should not have automatically gotten the update unless you are always using latest and for 0.x.x versions the minor version is treated like the major version. Also it was impossible to avoid a breaking change because the state variable went from being a Stream to a value so I thought it made more sense to make state/currentState a breaking change to make it clear that people had to do some renaming whereas dispose/dispatch are deprecated so you have until 1.0.0 to refactor those. Hope that helps 👍

larssn:
@felangel Thanks, I've sorted out our pubspec! 🙂

Any idea how we deal with the lint issue I mentioned above?

brianegan:
@felangel @larssn I think that's actually a bug with the linter package, although it might be a difficult one to fix. I'd recommend folks contribute examples to this issue: https://github.com/dart-lang/linter/issues/1381

Issue #561, opened 2019-10-10
Add Dart linters to the packages


tenhobi:
None of the packages uses lints in it's `analysis_options.yaml` file. (Except Angular which does have a few.)

My propose is to add more linters, so the code could easilly follow the same standards.

Maybe you know the `package:pedantic` Google internally uses as lints list. A few days ago I have made the `package:effective_dart` (https://github.com/tenhobi/effective_dart) which follows the Effecive Dart guide's lints, which I believe is suitable for our purpose.

Alternatively, we can make our own list, but the Effective Dart guide provides a good set of rule, I believe.

felangel:
Hi @tenhobi 👋 
Thanks for opening an issue!

I’ll definitely take a look at how effective_dart compares with pedantic and hopefully add one 👍. Thanks for bringing this up!

tenhobi:
@felangel cool, can I help you with something?

tenhobi:
Btw. also check https://github.com/dart-lang/linter/tree/master/example for more info. (btw. as for now, effective dart repo is also mentioned there) 🙂

felangel:
Awesome, I’ll try to take a look today 👍

tenhobi:
Another reason why to use lints according the effective dart guide is that you can now add this super cool badge to your collection. 🤣

[![follows: effective dart](https://img.shields.io/badge/follows-effective_dart-40c4ff.svg)](https://github.com/tenhobi/effective_dart)

(*just a joke ofc, consider what suits best for the project*)

felangel:
@tenhobi can you go over the differences between `effective_dart` and `pedantic`?

tenhobi:
Basically, `package:effective_dart` attempts to comply with all [Effective Dart](https://dart.dev/guides/language/effective-dart) guide rules which are described in that guide. `package:pedantic` sets only a very limited subset of those rules (partial implementation).

You can check the diff of https://github.com/tenhobi/effective_dart/blob/master/lib/analysis_options.1.0.0.yaml and https://github.com/dart-lang/pedantic/blob/master/lib/analysis_options.1.8.0.yaml.

The Effective Dart is the *ideal* state – the guide describes what you should do and don't do (and avoid, prefer, or concider) in your Dart project.

felangel:
@tenhobi thanks for the clarification 👍 

I want to evaluate what the implications (if any) would be tomorrow (flying back from vacation 😛) and then can quickly make the updates if everything looks good. Thanks!

[...]

PR #618, opened 2019-10-27
apply effective-dart style to flutter_bloc package

felangel:
## Status
**READY**

## Breaking Changes
NO

## Description
- Adhere to [Effective Dart](https://github.com/tenhobi/effective_dart) (#561)

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

None

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_builder.dart
+++ b/packages/flutter_bloc/lib/src/bloc_builder.dart
@@ -5,26 +5,26 @@ import 'package:flutter/widgets.dart';
 import 'package:bloc/bloc.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 
-/// Signature for the builder function which takes the [BuildContext] and state
-/// and is responsible for returning a [Widget] which is to be rendered.
-/// This is analogous to the `builder` function in [StreamBuilder].
+/// Signature for the [builder] function which takes the `BuildContext` and [state]
+/// and is responsible for returning a widget which is to be rendered.
+/// This is analogous to the [builder] function in [StreamBuilder].
 typedef BlocWidgetBuilder<S> = Widget Function(BuildContext context, S state);
 
-/// Signature for the condition function which takes the previous state and the current state
-/// and is responsible for returning a `bool` which determines whether or not to rebuild
-/// [BlocBuilder] with the current state.
+/// Signature for the condition function which takes the previous [state] and the current [state]
+/// and is responsible for returning a [bool] which determines whether or not to rebuild
+/// [BlocBuilder] with the current [state].
 typedef BlocBuilderCondition<S> = bool Function(S previous, S current);
 
 /// {@template blocbuilder}
-/// [BlocBuilder] handles building a widget in response to new states.
+/// [BlocBuilder] handles building a widget in response to new [states].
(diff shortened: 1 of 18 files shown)


felangel:
@tenhobi can you please review?

packages/flutter_bloc/lib/src/bloc_builder.dart
```diff
@@ -15,62 +15,65 @@ typedef BlocWidgetBuilder<S> = Widget Function(BuildContext context, S state);
...
+///   bloc: blocA,
+///   builder: (context, state) {
+///   // return widget here based on BlocA's state
+///   }
+/// )
+/// ```
+///
+/// An optional `condition` can be implemented for more granular control
+/// over how often [BlocBuilder] rebuilds.
+/// The `condition` function will be invoked on each bloc state change.
```
tenhobi: we can make `[condition]` etc. for fields like this

tenhobi: The same for all listener, builder etc. For all fields essentially.

felangel: Do you know what impact this will have on the generated docs? I'll run dart-doc and test it out

tenhobi: I think it will (only) allow to create more links between classes/methods/fields so user can navigate within the implementation in API reference page (https://pub.dev/documentation/bloc/latest/).

tenhobi: Basically, I think you should want to make a reference from everything you can, or that is how Flutter does it. And then the description is not just a plain text, but it has a context/references.

Check for example https://api.flutter.dev/flutter/widgets/Column-class.html

Dunno, what do you think?

tenhobi: It seems you can reference only methods, it's arguments and fields. But thats probably ok.

felangel: Yeah I just tested it locally and it was able to create references 👍 I just updated the PR can you take a look?

packages/flutter_bloc/lib/src/multi_repository_provider.dart
```diff
@@ -2,54 +2,57 @@ import 'package:flutter/widgets.dart';
...
+/// into a tree of nested [RepositoryProvider] widgets.
+/// As a result, the only advantage of using [MultiRepositoryProvider] is improved
+/// readability due to the reduction in nesting and boilerplate.
+/// {@endtemplate}
 class MultiRepositoryProvider extends StatelessWidget {
   /// The [RepositoryProvider] list which is converted into a tree of [RepositoryProvider] widgets.
   /// The tree of [RepositoryProvider] widgets is created in order meaning the first [RepositoryProvider]
   /// will be the top-most [RepositoryProvider] and the last [RepositoryProvider] will be a direct ancestor
-  /// of the `child` [Widget].
+  /// of the [child] [Widget].
```
tenhobi: Referencing widget don't feel right here.

felangel: good catch

jorgecoca:
Big fan of this PR! 🔝

Issue #611, opened 2019-10-24
[Proposal] Add blocTest to bloc_test


felangel:
**Is your feature request related to a problem? Please describe.**
When testing it's inconvenient to assert that a bloc emitted exactly a specific set of 
states and nothing more. Also, when expecting a state that the bloc never emits, we rely on the built-in 30s timeout which is not ideal.

**Describe the solution you'd like**
```dart
// PASSES
blocTest(
  'CounterBloc emits [0] when nothing is added',
  setUp: () => CounterBloc(),
  expect: [0],
);

// PASSES
blocTest(
  'CounterBloc emits [0, 1] when CounterEvent.increment is added',
  setUp: () => CounterBloc(),
  add: CounterEvent.increment,
  expect: [0, 1],
);

/*
FAILS IMMEDIATELY (no 30s timeout)
ERROR: Expected: [0]
  Actual: [0, 1]
   Which: longer than expected at location [1]
*/
blocTest(
  'CounterBloc emits [0] when CounterEvent.increment is added',
  setUp: () => CounterBloc(),
  add: CounterEvent.increment,
  expect: [0],
);

/*
FAILS IMMEDIATELY (no 30s timeout)
ERROR: Expected: [0, 1]
  Actual: [0]
   Which: shorter than expected at location [1]
*/
blocTest(
  'CounterBloc emits [0, 1] when no events are added',
  setUp: () => CounterBloc(),
  expect: [0, 1],
);
```

`blocTest` would ensure that the bloc emits the exact list of states without allowing subsequent states through.

**Additional context**
- Would address https://github.com/felangel/bloc/pull/564 without exposing a method which is only `visibleForTesting`.
- Would not require any changes to the core bloc api.

audkar:
Problem description looks good. I am just don't understand what solution is to solve this.

> Would also potentially require the close API to return a `Future` rather than be `void`.

Are you thinking to change behavior of existing `close` method? Currently it disposes pending events.

audkar:
@felangel can I provide any additional feedback?

felangel:
@audkar yes of course. I haven’t had much time to work on this and this proposal was just my initial thought. I would love your feedback 👍

audkar:
>  I haven’t had much time to work on this <...> I would love your feedback

Sorry bad wording. I saw GH tag `feedback wanted`, so I thought that this is blocked ticket to get additional feedback.

[...]

PR #626, opened 2019-10-29
Add blocTest

felangel:
## Status
**IN DEVELOPMENT**

## Breaking Changes
NO

## Description
- Add `blocTest` (addresses #611)

## Todos
- [X] Tests
- [X] Documentation
- [ ] Examples

None

### DIFF
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -1,0 +1,80 @@
+import 'dart:async';
+
+import 'package:bloc/bloc.dart';
+import 'package:meta/meta.dart';
+import 'package:test/test.dart';
+
+/// Creates a new [bloc]-specific test case with the given [description].
+/// [blocTest] will handle asserting that the [bloc] emits the [expect]ed
+/// states (in order) after the given event is [add]ed.
+/// [blocTest] also handles ensuring that no additional states are emitted
+/// by closing the bloc stream before evaluating the [expect]ation.
+///
+/// [setUp] should be used to all [bloc] initialization and preparation
+/// and must return the [bloc] under test.
+///
+/// [add] is an optional event which will be [add]ed to the [bloc].
+///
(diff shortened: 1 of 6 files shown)


audkar:
Tests pass only because they doesn't do any async work while handling event. Real world example would be `CounterBloc.dart`:
```dart 
  @override
  Stream<int> mapEventToState(
    CounterEvent event,
  ) async* {
    switch (event) {
      case CounterEvent.increment:
        await Future.delayed(Duration(microseconds: 1)); //<------------------ with this
        yield state + 1;
        break;
    }
  }
```

felangel:
@audkar yup I’m aware and am working on that today 👍

packages/bloc_test/lib/src/bloc_test.dart
```diff
@@ -0,0 +1,80 @@
...
+///
+/// [add] is an optional event which will be [add]ed to the [bloc].
+///
+/// [expect] is an `Iterable<State>` which we expect the [bloc]
+/// under test to emit after the provided event is [add]ed.
+///
+/// ```dart
+/// blocTest(
+///   'CounterBloc emits [0, 1] when CounterEvent.increment is added',
+///   setUp: () => CounterBloc(),
```
jorgecoca: Maybe `build/builder` instead of `setUp`? I "think" it might lead to confusion with the preexisting `setUp` method of the test suite...

packages/bloc_test/lib/src/bloc_test.dart
```diff
@@ -0,0 +1,80 @@
...
+/// ```
+@isTest
+void blocTest<B extends Bloc<Event, State>, Event, State>(
+  String description, {
+  @required B setUp(),
+  @required Iterable<State> expect,
+  Event add,
+}) {
+  test(description, () async {
+    final bloc = setUp();
```
jorgecoca: Do you think is worth it to check for nullables?

PR #629, opened 2019-10-30
Example of flutter_bloc with navigation

RollyPeres:
## Status
READY

## Breaking Changes
NO

## Description
Provides an example of using flutter_bloc to share a bloc between multiple routes when routing with `.pushNamed`.

## Todos
- [ ] Tests
- [x] Documentation
- [ ] Examples


tenhobi:
Wouldn't it be better to add the process of using named routes to an existing recipe covering accessing bloc https://bloclibrary.dev/#/recipesflutterblocaccess?

As a user, I would like to see the information in one place. Could you add a simple recipe there as well?

---

Note that this comment does not force you not to make this tutorial about navigation using BLoC. It might also be very useful if done properly.

Will you make a markdown tutorial out of this as well, right? (like e.g. https://bloclibrary.dev/#/flutterlogintutorial)

RollyPeres:
@tenhobi It would probably be better. Unfortunately, I'm really busy, I don't have the time to do what you're asking for. It's up to you guys what you do with this PR ✌

felangel:
I agree with @tenhobi that this should ideally be incorporated into the blocaccess recipe. Since you're busy @RollyPeres I'll close this PR and try to adapt parts of your solution into an update to the existing recipe 👍 

Thanks for putting this together!

PR #638, opened 2019-11-03
update bloc.close to process pending events

felangel:
## Status
**READY**

## Breaking Changes
YES

## Description
Previously calling `close` on a bloc would immediately ignore all pending events. This pull request would change the behavior so that calling `close` would prevent new events from being processed while allowing pending events to finish being processed (see conversation in #611 and #639 for more details).

## Todos
- [X] Tests
- [x] Documentation
- [x] Examples

- Breaking change in bloc

### DIFF
--- a/packages/bloc/README.md
+++ b/packages/bloc/README.md
@@ -55,7 +55,7 @@ This design pattern helps to separate _presentation_ from _business logic_. Foll
 
 **onError** is a method that can be overridden to handle whenever an `Exception` is thrown. By default all exceptions will be ignored and `Bloc` functionality will be unaffected. **It is a great place to add bloc-specific error handling**.
 
-**close** is a method that closes the `event` and `state` streams. `close` should be called when a `Bloc` is no longer needed. Once `close` is called, `events` that are `added` will not be processed and will result in an error being passed to `onError`. In addition, if `close` is called while `events` are still being processed, any `states` yielded after are ignored and will not result in a `Transition`.
+**close** is a method that closes the `event` and `state` streams. `close` should be called when a `Bloc` is no longer needed. Once `close` is called, `events` that are `added` will not be processed and will result in an error being passed to `onError`. In addition, if `close` is called while `events` are still being processed the `bloc` will continue to process the pending `events` to completion.
 
 ## BlocDelegate Interface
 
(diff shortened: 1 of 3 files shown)


packages/bloc/README.md
```diff
@@ -55,7 +55,7 @@ This design pattern helps to separate _presentation_ from _business logic_. Foll
 
 **onError** is a method that can be overridden to handle whenever an `Exception` is thrown. By default all exceptions will be ignored and `Bloc` functionality will be unaffected. **It is a great place to add bloc-specific error handling**.
 
-**close** is a method that closes the `event` and `state` streams. `close` should be called when a `Bloc` is no longer needed. Once `close` is called, `events` that are `added` will not be processed and will result in an error being passed to `onError`. In addition, if `close` is called while `events` are still being processed, any `states` yielded after are ignored and will not result in a `Transition`.
+**close** is a method that closes the `event` and `state` streams. `close` should be called when a `Bloc` is no longer needed. Once `close` is called, `events` that are `added` will not be processed and will result in an error being passed to `onError`. In addition, if `close` is called while `events` are still being processed the `bloc` will continue to process the pending `events` to completion.
```
tenhobi: Will this be changed also in docs?

felangel: Yes the english docs reference this file directly so we'll need to update the translated READMEs as well

tenhobi: I assume by looking at docs sources that it only *references* (link) this README, right? So it's not a part of translations.

Also, there are many places where is used `void closed()` used, and that should be changed to `Future<void> close() async {}`, if I am correct.

And, please, could you create some way to let us know when some docs are updated, so we can keep up with changes? Maybe create some issue for this and each new doc PR will mention the issue so it gets the notification? Or mention us on every PR?

felangel: Yup, you're correct. We'll need to figure out a good process. For now I can create a separate project for translations and create tickets every time the documentation is changed. Thoughts?

felangel: I'm also wondering if services like [weglot](https://weglot.com) would be a good solution to automating the translations. That way you and other members of the community can fix bugs or make improvements to the translations as needed but the bulk of the work would be automated.

PR #645, opened 2019-11-04
bloc v2.0.0

felangel:
## Status
**READY**

## Breaking Changes
NO

## Description
- bloc v2.0.0

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

- new major version of bloc

### DIFF
--- a/packages/bloc/CHANGELOG.md
+++ b/packages/bloc/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 2.0.0
+
+- Allow blocs to finish processing pending events on `close` ([#639](https://github.com/felangel/bloc/issues/639))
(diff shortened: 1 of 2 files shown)


tenhobi:
As mentioned at https://github.com/felangel/bloc/pull/638/files#r342288000, the docs should probably update the `close` method overriding etc?

felangel:
> As mentioned at https://github.com/felangel/bloc/pull/638/files#r342288000, the docs should probably update the `close` method overriding, etc.? Or can we still override it as `void close()` while its `Future<void> close async {}` in Bloc?
> 
> image

I'll need to do a round of documentation updates before publishing and I will create a ticket to update the translations as well

PR #671, opened 2019-11-14
previousState updating

Ramesh-X:
## Status
READY

## Breaking Changes
NO

## Description
Currently `BlocBuilder` is changing its `previousState` no matter the `Widget` was built or not. But this PR will update `previousState` only if the `Widget` was rebuilt.

This might be helpful when number of streams do changes to the `State` in bloc, but you need to rebuild `Widget` only if the all the changes are correctly done.

## Todos
- [x] Tests
- [x] Documentation
- [x] Examples

This PR will affect:

* `BlocBuilder` build logic is changed

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_builder.dart
+++ b/packages/flutter_bloc/lib/src/bloc_builder.dart
@@ -167,8 +167,8 @@ class _BlocBuilderBaseState<B extends Bloc<dynamic, S>, S>
           setState(() {
             _state = state;
           });
+          _previousState = state;
         }
-        _previousState = state;
       });
     }
   }


felangel:
@Ramesh-X thanks for opening a pull request. Can you please provide some more context? Is there a concrete use-case that you have?

Ramesh-X:
Consider the following case where I have two actions called `F1Action` and `F2Action` that will be dispatched asynchronously. The `BlocBuilder` will be rebuilt on every state change only if the both variables have changed.

```dart
  class A {
    int f1;
    int f2;
    A(this.f1, this.f2);
  }

  @override
  Stream<A> mapEventToState(RootPageAction action) async* {
    if (action is F1Action) {
      yield A(action.val, state.f2);
    } else if (action is F2Action) {
      yield A(state.f1, action.val);
    }
  }

  return BlocBuilder<MyBloc, A>(
    condition: (pre, current) => pre.f1 != current.f1 && pre.f2 != current.f2,
    builder: ...
  );
```

With the existing code, the `BlocBuilder` will never be rebuilt, because in every state change either `f1` or `f2` will remain constant. Therefore the condition will not be satisfied every time.

But with my PR, I'm keeping the previous state which the `BlocBuilder` was rebuilt and compare the current state with that.

Despite of the example, I think it is better to do it this way anyhow. Because a `Widget` should always be rebuilt only if the state it was previously built has changed (and satisfy the condition). It should not consider the state changes that was not contributed to any rebuild. I hope you can understand the point that I'm making..

You could tell that I can satisfy my requirements by changing `&&` to `||`. Of course it will do the thing I want but with a few unnecessary rebuilds.

felangel:
@Ramesh-X thanks for the clarification 👍. For the PR to be able to be merged you’ll need to fix unit tests. Also, a similar change should also be made to BlocListener. Are you able to make those changes or would you like me to? Thanks again!

felangel:
@Ramesh-X I've opened a PR with the improvements to both `BlocListener` and `BlocBuilder` at https://github.com/felangel/bloc/pull/678

Ramesh-X:
I will add necessary tests and push..

PR #678, opened 2019-11-17
[Patch] flutter_bloc condition (previous state)

felangel:
## Status
**READY**

## Breaking Changes
NO

## Description
- Patch `BlocBuilder` and `BlocListener` condition to be called with the previous used state of the bloc rather than the previous state of the bloc (#671)

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,6 @@ pubspec.lock
 coverage/
 .test_coverage.dart
 *.log
+flutter_export_environment.sh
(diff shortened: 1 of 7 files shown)


ezamagni:
This was a breaking change:
Any `BlocBuilder` / `BlocListener` with a `condition` on `prevState` would much likely have its behavior changed. See #709

Issue #545, opened 2019-09-30
Combined BlocBuilder/Listener


fl0cke:
**Is your feature request related to a problem? Please describe.**
I often find myself in the situation where I need both bloc builders and bloc listeners in the same widget tree, e.g. for showing a loading indicator (builder) and displaying a snackbar on success (listener).

**Describe the solution you'd like**
A BlocBuilderListener Widget that accepts both a builder function and a listener would greatly reduce the clutter in my code and also reduce the overhead of having two stateful widgets that basically do the same. Alternatively, you could add a listener parameter to the existing BlocBuilder widget.

felangel:
Hi @fl0cke 👋 
Thanks for opening an issue!

Can you please elaborate a bit on why you say this would reduce the clutter in your code?
If I'm understanding correctly, you're proposing to go from:

```dart
BlocListener<MyBloc, MyState>(
  listener: (context, state) { ... },
  child: BlocBuilder<MyBloc, MyState>(
    builder: (context, state) { ... }
  ),
),
```

to

```dart
BlocBuilderListener<MyBloc, MyState>(
  listener: (context, state) { ... },
  builder: (context, state) { ... },
),
```

I personally don't think it's cluttered as is and I'd be more concerned about confusing people because it isn't immediately obvious what `BlocBuilderListener` would do.

Thoughts?

fl0cke:
Hey @felangel,
Thanks for the quick reply!

Yes, that's exactly what the proposed BlocBuilderListener would look like. Alternatively, we could add an optional `listener` parameter to the `BlocBuilder` to avoid introducing a new widget. 

I agree that it's not that much clutter, but when i look at my usage of the two BlocX widgets, i often find myself using either just a `BlocBuilder` or a `BlocBuilder` nested in a `BlocListener`, but almost never a `BlocListener` on its own. So this would be good opportunity to reduce the number of widgets that are required to achieve a reactive UI + side effects.

Example: A common use case is submitting some sort of user input and displaying a loading indicator while doing so. On success, I want to both show a SnackBar (using the listener) and hide the loading indicator (using the builder).

bigworld12:
I would suggest the merge to have the same signature as `ListView` and `ListView.builder`

e.g.

```dart
//normal BlocListener, listener is triggered on state change, child doesn't get rebuilt on state change
BlocView<MyBloc,MyState>(
  listener: (context, state) { ... }, 
  child: Container()
)

[...]

PR #753, opened 2019-12-31
Add BlocConsumer

felangel:
## Status
**READY**

## Breaking Changes
NO

## Description
Add `BlocConsumer` to `flutter_bloc` (https://github.com/felangel/bloc/issues/545)

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

- New Feature which will be included in `flutter_bloc v3.1.0`

### DIFF
--- a/docs/flutterbloccoreconcepts.md
+++ b/docs/flutterbloccoreconcepts.md
@@ -192,6 +192,45 @@ MultiBlocListener(
 )
 ```
 
+### BlocConsumer
+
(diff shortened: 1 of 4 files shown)


docs/flutterbloccoreconcepts.md
```diff
@@ -192,6 +192,45 @@ MultiBlocListener(
 )
 ```
 
+### BlocConsumer
+
+**BlocConsumer** exposes a `builder` and `listener` in order react to new states. `BlocConsumer` is analogous to a nested `BlocListener` and `BlocBuilder` but reduces the amount of boilerplate needed. `BlocConsumer` should only be used when it is necessary to both rebuild UI and execute other reactions to state changes in the `bloc`. `BlocConsumer` takes a required `BlocWidgetBuilder` and `BlocWidgetListener` and an optional `bloc`, `BlocBuilderCondition`, and `BlocListenerCondition`.
+
+If the `bloc` parameter is omitted, `BlocConsumer` will automatically perform a lookup using
```
anticafe: Could you add more detail about `perform a lookup` for what? I think it is `perform a lookup for the BlocA instance`.

Issue #724, opened 2019-12-13
blocTest does not catch all emitted states


MohsinN:
`bloc_test` does not catch states emitted by consecutive events added inside bloc. Same scenario passes in plain test but fails with `blocTest`.

I updated your `flutter_login` sample to replicate the issue.

Updated `AuthenticationBloc` to add  `LoggedOut` event inside case of  `LoggedIn`

```dart
import 'dart:async';

import 'package:bloc/bloc.dart';
import 'package:flutter_login/authentication/authentication.dart';
import 'package:meta/meta.dart';
import 'package:user_repository/user_repository.dart';

class AuthenticationBloc
    extends Bloc<AuthenticationEvent, AuthenticationState> {
  final UserRepository userRepository;

  AuthenticationBloc({@required this.userRepository})
      : assert(userRepository != null);

  @override
  AuthenticationState get initialState => AuthenticationUninitialized();

  @override
  Stream<AuthenticationState> mapEventToState(
    AuthenticationEvent event,
  ) async* {
    if (event is AppStarted) {
      final bool hasToken = await userRepository.hasToken();

      if (hasToken) {
        yield AuthenticationAuthenticated();
      } else {
        yield AuthenticationUnauthenticated();
      }
    }

    if (event is LoggedIn) {
      yield AuthenticationLoading();
      await userRepository.persistToken(event.token);
      yield AuthenticationAuthenticated();
      add(LoggedOut());
    }

    if (event is LoggedOut) {
      yield AuthenticationLoading();
      await userRepository.deleteToken();
      yield AuthenticationUnauthenticated();
    }
  }
}
```
In the first case, only first three states are caught by `blocTest`
Updated tests:
```dart
blocTest(
        ' Failing test: emits [uninitialized, loading, authenticated, loading, unauthenticated] when token is persisted',
        build: () => authenticationBloc,
        act: (bloc) => bloc.add(LoggedIn(
              token: 'instance.token',
            )),
        expect: [
          AuthenticationUninitialized(),
          AuthenticationLoading(),
          AuthenticationAuthenticated(),
          AuthenticationLoading(),
          AuthenticationUnauthenticated(),
        ]);

    test(
        'emits [uninitialized, loading, authenticated, loading, unauthenticated] when token is persisted',
        () {
      final expectedResponse = [
        AuthenticationUninitialized(),
        AuthenticationLoading(),
        AuthenticationAuthenticated(),
        AuthenticationLoading(),
        AuthenticationUnauthenticated(),
[...]

PR #756, opened 2020-01-02
allow blocTests to add more than one asynchronous event

dustin-graham:
## Status
READY

## Breaking Changes
NO

## Description
allow blocTests to add more than one asynchronous event

The issue was that emitsExactly was listening to events too late. Since
Bloc uses a BehaviorSubject, if there is a delay between events the
listener would miss states emitted before the listener attaches. The
listener should start listening before any events are added so that all
states get recorded. I suspect this was the underlying problem with
[#724].
## Todos
- [x] Tests
- [x] Documentation
- [x] Examples

## Steps to Test or Reproduce
The included test outlines the existing problem. If there is an asynchronous situation and the test requires multiple events in sequence, the current implementation will yield a false negative as it may not capture all the bloc's states. The reason is that act is called before the emitExactly function listens to the bloc. The behaviorSubject in the bloc starts with the last state given which may exclude earlier states in some situations. By listening to the bloc stream before we act, this should solve the problem.

This PR will affect: nothing else as far as I can tell. all the bloc_test tests pass with this change.

### DIFF
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -78,7 +78,6 @@ void blocTest<B extends Bloc<Event, State>, Event, State>(
 }) {
   test(description, () async {
     final bloc = build();
-    await act?.call(bloc);
-    await emitsExactly(bloc, expect, duration: wait);
+    await emitsExactly(bloc, expect, duration: wait, act: act);
   });
 }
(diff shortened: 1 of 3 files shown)


felangel:
Hi @dustin-graham 👋 
Thanks for opening this PR. I thought about doing this but decided against it because it makes tightly couples the `blocTest` and `emitsExactly` API. I would recommend closing this PR and I can open a separate PR to address this in `blocTest` without modifying the `emitsExactly` API. Let me know what you think and thanks again!

dustin-graham:
Thanks! That would be fine. But perhaps you could just modify my pull
request with the needed changes. That way, once done I can readily run my
tests with the changes to see if that works for my user case? Either way.

felangel:
@dustin-graham can you verify https://github.com/felangel/bloc/pull/760? I kept the test you added and included several others as well 👍

PR #768, opened 2020-01-07
Upgrade to effective_dart 1.2.0

tenhobi:
## Status
**DONE**

## Breaking Changes
NO

## Description

Hi, as we discussed in the chat, I have upgraded `package:effective_dart` to `1.2.0` in `packages/`. It turns out that the major updates needed are relative imports within your package, https://dart-lang.github.io/linter/lints/avoid_types_on_closure_parameters.html and 80 chars in line (but that mostly for comments).  

Also, see https://github.com/tenhobi/effective_dart/blob/master/CHANGELOG.md

I did updated everything except comments. Can you tell me how (in what way/style) should I update them? E.g. where to choose the breaks. :slight_smile: And also https://dart-lang.github.io/linter/lints/avoid_equals_and_hash_code_on_mutable_classes.html, because that requires maybe some more look. (it's avoid btw, so you can opt out it).

For example, should

```
/// Called whenever an [event] is `added` to any [bloc] with the given [bloc] and [event].
/// A great spot to add universal logging/analytics.
```

become

```
/// Called whenever an [event] is `added` to any [bloc]
/// with the given [bloc] and [event].
/// A great spot to add universal logging/analytics.
```

or

```
/// Called whenever an [event] is `added` to any [bloc]
/// with the given [bloc] and [event]. A great spot to
/// add universal logging/analytics.
```

---

Also, I probably need to play a bit also with analyzer, so it is cleer what is prefer/avoid (hint) and do/don't (warning). But for that I will need more help/know-how. Will take a look in weekend. :slight_smile:  

## Todos

- [x] info: AVOID lines longer than 80 characters. (lines_longer_than_80_chars)
- [x] info: AVOID overloading operator == and hashCode on classes not marked `@immutable`. (avoid_equals_and_hash_code_on_mutable_classes)

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -31,8 +31,9 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
   }
 
   /// Adds a subscription to the `Stream<State>`.
-  /// Returns a [StreamSubscription] which handles events from the `Stream<State>`
-  /// using the provided [onData], [onError] and [onDone] handlers.
+  /// Returns a [StreamSubscription] which handles events from
+  /// the `Stream<State>` using the provided [onData], [onError] and [onDone]
+  /// handlers.
   @override
   StreamSubscription<State> listen(
     void Function(State) onData, {
@@ -53,14 +54,17 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
   void onEvent(Event event) => null;
 
   /// Called whenever a [transition] occurs with the given [transition].
-  /// A [transition] occurs when a new `event` is [add]ed and [mapEventToState] executed.
+  /// A [transition] occurs when a new `event` is [add]ed and [mapEventToState]
+  /// executed.
   /// [onTransition] is called before a [bloc]'s [state] has been updated.
   /// A great spot to add logging/analytics at the individual [bloc] level.
   void onTransition(Transition<Event, State> transition) => null;
 
   /// Called whenever an [error] is thrown within [mapEventToState].
-  /// By default all [error]s will be ignored and [bloc] functionality will be unaffected.
-  /// The [stacktrace] argument may be `null` if the [state] stream received an error without a [stacktrace].
+  /// By default all [error]s will be ignored and [bloc] functionality will be
+  /// unaffected.
+  /// The [stacktrace] argument may be `null` if the [state] stream received
+  /// an error without a [stacktrace].
   /// A great spot to handle errors at the individual [Bloc] level.
   void onError(Object error, StackTrace stacktrace) => null;
 
@@ -83,7 +87,8 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
   /// This method should be called when a [bloc] is no longer needed.
   /// Once [close] is called, `events` that are [add]ed will not be
   /// processed and will result in an error being passed to [onError].
-  /// In addition, if [close] is called while `events` are still being processed,
+  /// In addition, if [close] is called while `events` are still being
+  /// processed,
   /// the [bloc] will continue to process the pending `events` to completion.
   @override
   @mustCallSuper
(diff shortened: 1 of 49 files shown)


tenhobi:
// Now it only fails because of the listed issues.

felangel:
Thanks for the contribution! 💯 
I would prefer to break on new sentences personally for the inline documentation. Let me know if you need any help and thanks again 😄

packages/bloc/lib/src/bloc.dart
```diff
@@ -1,9 +1,10 @@
 import 'dart:async';
 
-import 'package:bloc/bloc.dart';
 import 'package:meta/meta.dart';
 import 'package:rxdart/rxdart.dart';
 
+import '../bloc.dart';
```
felangel: I thought we should prefer package imports for all. What rule are we breaking here?

packages/bloc/test/bloc_delegate_test.dart
```diff
@@ -2,7 +2,7 @@ import 'package:bloc/bloc.dart';
 import 'package:mockito/mockito.dart';
 import 'package:test/test.dart';
 
-import './helpers/helpers.dart';
+import 'helpers/helpers.dart';
```
felangel: 💯

packages/flutter_bloc/test/bloc_builder_test.dart
```diff
@@ -40,8 +39,8 @@ class MyThemeAppState extends State<MyThemeApp> {
     return BlocBuilder(
       bloc: _themeBloc,
       builder: ((
-        BuildContext context,
-        ThemeData theme,
+        context,
+        theme,
```
felangel: minor but might read better without the trailing comma after `theme`

packages/bloc/lib/src/bloc.dart
tenhobi: https://dart.dev/guides/language/effective-dart/usage#prefer-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory

felangel: Thanks, I'm still a bit confused because I was under the impression that the rule was only if you are referencing internal members (`./lib/src/*`) however in this case we are referencing public members of the library.

tenhobi: As I understand it, it says that:

- you should never use `../` outside of lib and have `lib/` in your path
- you cannot reference from outside of `lib/` (`bin/`, `test/`, ...) with relative path
- in `lib/` prefer to use relative paths

it doesn't say anything about `lin/src/*` I believe. Maybe this sentence "Libraries inside lib can import other libraries inside lib (or in subdirectories of it)." is what confuses you?

This lint rule was implemented by https://github.com/dart-lang/linter/pull/1738. There are links to 2 issues where one of that, mine, references the specs from the Effective Dart. So it should be valid. :-)

tenhobi:
> Thanks for the contribution! 💯
> I would prefer to break on new sentences personally for the inline documentation. Let me know if you need any help and thanks again 😄

So you prefer this, right?

```dart
/// Called whenever an [event] is `added` to any [bloc]
/// with the given [bloc] and [event].
/// A great spot to add universal logging/analytics.
```

---

I will upgrade the rest tomorrow. :-)

packages/bloc/lib/src/bloc.dart
tenhobi: And it depends on you if you want to reference the public library file. I would personally opt for referencing only inside `lib/src/*`, as it probably makes the code a bit cleaner.

tenhobi:
Btw. from this guide https://github.com/dart-lang/site-shared/wiki/Writing-for-Dart-and-Flutter-websites I have discovered an amazing https://sembr.org/ (Semantic Line Breaks) :-)

packages/bloc/lib/src/bloc.dart
tenhobi: @felangel should I keep it this way, or avoid importing a public library and therefore import other libraries directly? ;)

tenhobi:
I will need your help with all of `info: AVOID overloading operator == and hashCode on classes not marked @immutable. (avoid_equals_and_hash_code_on_mutable_classes)`, some cases might be fixed by adding that annotation, but some don't. We can add ignore comments for the parts where it has no meaning. It's only "avoid" rule, not "don't".

felangel:
> I will need your help with all of `info: AVOID overloading operator == and hashCode on classes not marked @immutable. (avoid_equals_and_hash_code_on_mutable_classes)`, some cases might be fixed by adding that annotation, but some don't. We can add ignore comments for the parts where it has no meaning. It's only "avoid" rule, not "don't".

Opened https://github.com/tenhobi/bloc/pull/1 to fix the remaining errors 👍

packages/bloc/lib/src/bloc.dart
felangel: I'm fine with keeping it as is

PR #777, opened 2020-01-13
Add GitHub CI

tenhobi:
## Status
**READY**

## Breaking Changes
NO

## Description
Proposal to migrate from Travis CI to GitHub Actions.

I am not sure how the current way with uploading to codecov works. ~This solution expect that you add secrets to the repository (`CODECOV_TOKEN_flutter_bloc` etc.).~

Also, not sure about the path for caching Flutter.

Some resources:
- https://github.com/actions/cache
- https://github.com/subosito/flutter-action
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#add-a-system-path-add-path
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#configuring-a-build-matrix
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

## Todos
- [x] Test if that works (see https://github.com/tenhobi/bloc/actions)
- [x] Codecov (`CODECOV_TOKEN`)
- [x] Caching Flutter
- [x] Remove Travis CI
- [x] Update badges etc.

If you wanna change CI to build or whatever, change to `name: CI` is required.
`[Actions status](https://github.com/felangel/bloc/actions)`

Preview:
[Actions status](https://github.com/tenhobi/bloc/actions) (this is for my forked repo)


.github/workflows/main.yaml
```diff
@@ -0,0 +1,39 @@
...
+    strategy:
+      matrix:
+        package: ['angular_bloc', 'bloc', 'bloc_test', 'flutter_bloc']
+      fail-fast: true
+
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '12.x'
```
felangel: why do we need `setup-java`?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,39 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    container:
+      image: google/dart:latest
```
felangel: I would prefer to use Flutter stable and the corresponding Dart version

tenhobi: Well for dart_test and angular_bloc you shouldn't use Flutter SDK, ay? So for that, I have included Dart also.

.github/workflows/main.yaml
tenhobi: https://github.com/subosito/flutter-action/issues/15 this says it's somehow broken otherwise? dunno, we might try to remove it.

.github/workflows/main.yaml
felangel: yeah I agree that we need the Dart SDK but would like to use whatever version of Dart the latest stable version of Flutter is using.

.github/workflows/main.yaml
felangel: If it's not too much trouble can you try without it? Seems weird to me 🤔

.github/workflows/main.yaml
tenhobi: Hmm. Could we use flutter too for that? `flutter pub get` etc should work in angular and bloc_test as well?

.github/workflows/main.yaml
tenhobi: Will try.

tenhobi:
Finally, it seems to work – https://github.com/tenhobi/bloc/runs/387793998

@felangel could you add a `CODECOV_TOKEN` to this repo?

tenhobi:
Now it also runs for macOS https://github.com/tenhobi/bloc/runs/387835505 💪

tenhobi:
It also takes **much less** (+- 3 times) to process.

image
image

tenhobi:
I am doing some experiments with caching etc. Changing status to **in progress**.

tenhobi:
I am investigating that macOS doesn't do caching for some reason, Ubuntu works well. Not a major issue tho, it's 30s vs 1m 20s. Can be fixed later.

More details here: https://github.com/subosito/flutter-action/issues/16

tenhobi:
It was probably lost in our DM, so I am writing it here. From that linked issue, the guy had a good point. We should have a version in the key because otherwise it would not (probably?) update the data end we end up updating every time after a new version is published.

So, we have two options

1) use a version we provide, which might be good because you control what version your using and that might prevent some issues (and I think this is good for app development, but not sure about our case)
2) or make some script which would fetch the latest version of flutter and somehow past that

felangel:
@tenhobi is this ready for review now or are you still working on the MacOS caching?

tenhobi:
The only missing thing is in my last message. :) You should choose.

felangel:
Ah sorry I missed it! I think we can just fix it to the latest stable version of flutter and update as flutter is updated.

tenhobi:
I have added that specific version. Awaiting passing the CI.

Dunno if this is due to the latest changes, but the macOS tasks are way toooooooooooooooo slow. :O

tenhobi:
@felangel ok, it was probably some slow down yesterday. I think we are ready to merge.

PR #821, opened 2020-01-30
Remove RxDart dependency from bloc package

mmozhaiskyi:
## Status
**READY**

## Breaking Changes
NO

## Description
BloC is architecture library which used as the foundation of the application. In my opinion it should be independent of any other libraries like [RxDart](https://pub.dev/packages/rxdart).

Rx code was used only in two places inside bloc package.

* *bloc.dart* (`PublishSubject`, `BehaviorSubject` replaced with two `StreamController`. Also added variable for storing current state which called `_currentState`)
* *complex_bloc.dart* (File is test helper. It still using `RxDart` because package is enable as a test dependency)


packages/bloc/lib/src/bloc.dart
```diff
@@ -10,23 +9,26 @@ import '../bloc.dart';
...
-  State get state => _stateSubject.value;
+  State get state => _currentState;
 
   /// Returns the [state] before any `events` have been [add]ed.
   State get initialState;
 
   /// Returns whether the `Stream<State>` is a broadcast stream.
   @override
-  bool get isBroadcast => _stateSubject.isBroadcast;
+  bool get isBroadcast => _states.stream.isBroadcast
```
felangel: I believe this should just be `_states.stream.isBroadcast` since this method is overridden for `Stream`.

packages/bloc/lib/src/bloc.dart
```diff
@@ -10,23 +9,26 @@ import '../bloc.dart';
 /// and transforms them into a `Stream` of `States` as output.
 /// {@endtemplate}
 abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
-  final PublishSubject<Event> _eventSubject = PublishSubject<Event>();
+  final _states = StreamController<State>();
+  final _events = StreamController<Event>.broadcast();
```
felangel: Just curious, why does this need to broadcast?

tenhobi:
I am not sure atm, but using `BehaviorSubject` you are given the last emitted state (from cache) when you subscribe (was `_stateSubject.listen`) even after some changes to the bloc have been made (emitting a new state). What part of the code does this now? Not sure it's part of the tests to cover this issue.

> BehaviorSubject - A broadcast StreamController that caches the latest added value or error. When a new listener subscribes to the Stream, the latest value or error will be emitted to the listener. Furthermore, you can synchronously read the last emitted value.

What I mean:

1. subscribe to BlocA with X
1. add an event1
1. that emits a state1
1. subscribe to BlocA with Y
1. does the Y gets state1 now ??

jorgecoca:
> I am not sure atm, but using `BehaviorSubject` you are given the last emitted state (from cache) when you subscribe (was `_stateSubject.listen`) even after some changes to the bloc have been made (emitting a new state). What part of the code does this now? Not sure it's part of the tests to cover this issue.
> 
> > BehaviorSubject - A broadcast StreamController that caches the latest added value or error. When a new listener subscribes to the Stream, the latest value or error will be emitted to the listener. Furthermore, you can synchronously read the last emitted value.
> 
> What I mean:
> 
> 1. subscribe to BlocA with X
> 2. add an event1
> 3. that emits a state1
> 4. subscribe to BlocA with Y
> 5. does the Y gets state1 now ??

@felangel is there a test case for this? If not, I think that would be a good addition ;)

packages/bloc/pubspec.yaml
```diff
@@ -10,10 +10,11 @@ environment:
...
 dependencies:
   meta: ^1.1.6
-  rxdart: ^0.23.0
 
 dev_dependencies:
   test: ">=1.3.0 <2.0.0"
   test_coverage: ^0.2.0
   mockito: ^4.0.0
   effective_dart: ^1.2.0
+  rxdart: ^0.23.0
```
jorgecoca: Is there a reason to keep this as a `dev_dependency`? I think it might be worth to totally remove `rxdart` if possible!

packages/bloc/lib/src/bloc.dart
jorgecoca: I'm confused by this as well...

packages/bloc/pubspec.yaml
felangel: There are some tests that use rx operators (debounceTime, etc...)

packages/bloc/lib/src/bloc.dart
mmozhaiskyi: Thanks. Removed unnecessary broadcast stream. It effects `triggers onError from add` from `bloc_test.dart` file. Looks like because broadcast and single subscription streams contains different error message which it test expect. It is funny.

Broadcast stream: `Cannot add new events after calling close`
Single subscription stream: `Cannot add event after closing`

packages/bloc/lib/src/bloc.dart
mmozhaiskyi: Hmm. Looks like `isBroadcast` value always equal `true` in previous version because `BehaviorSubject` contains `StreamController.broadcast()` as implementation which always return `isBroadcast =>true`. I changed this line to `isBroadcast => true`. @felangel What do you think?

felangel: hmm I need to take a closer look at the BehaviorSubject and PublishSubject implementations but I would try to stay as consistent as possible with the existing implementations. I would also prefer to return `_state.stream.isBroadcast` rather than hardcoding the value in case the underlying implementation changes at some point. Thoughts?

packages/bloc/pubspec.yaml
jorgecoca: gotcha!

mmozhaiskyi:
PR contains problems. Changed status to `IN DEVELOPMENT`. I will try to find better solution.

@tenhobi you are right. Solution with `StreamController` does not include previous value when new subscription established. It is critical (missing initial state when widget was rebuilded). Also `_states` should be a broadcast to allow multi subscriptions.

I propose to add test to cover this issue in separate PR. It will be very useful.
>1. subscribe to BlocA with X
>2. add an event1
>3. that emits a state1
>4. subscribe to BlocA with Y
>5. does the Y gets state1 now ??

@tenhobi thank you. Good catch 👍

mmozhaiskyi:
Implemented `StateSubject` (please propose different name if it is necessary). It is lightweight analog of `BehaviorSubject`. This approach fixed issues related to last item emit. Also BLoC support several listeners as well.

Added tests which cover `StateSubject` logic such as `emits`, `seed item emit`, `errors`, `null-values`. Test cases with several listeners are included.

FYI @felangel @tenhobi

felangel:
@mmozhaiskyi thanks for your efforts! I have a few concerns about this approach and would love to hear your thoughts:

1. `StateSubject` should not be publicly exposed in my opinion as it's just an implementation detail.
2. I don't like the idea of copying the `BehaviorSubject` implementation straight from `rxdart` just to remove the dependency. It's not a trivial implementation (especially in it's current form) and I personally feel like it could be even further simplified because it will be used in a very specific context.
3. Also, we now run the risk of having to maintain the same implementation in multiple places which isn't beneficial for the community and doesn't feel right to me.

Thoughts? (cc @brianegan)

felangel:
Was just messing around on [my own branch](https://github.com/felangel/bloc/pull/838) and managed to simplify the implementation a bit more (I think). What do y'all think?

mmozhaiskyi:
@felangel Thank you for your feedback.

[Your proposal](https://github.com/felangel/bloc/pull/838) is awesome and useful. Really, we don’t need full implementation of `BehaviourSubject` from Rx because bloc context is very specific. I simplified solution as it possible (see last commit). 

Changes:
* current approach uses  `_StartWithStreamTransformer` only. 
* logic of last value emit moved into bloc (StateSubject abstraction was removed).

brianegan:
My take: I definitely think Bloc can be a fine library without Rx, as Felix said, especially with the changes to extension methods. Folks can opt-in when they need RxDart much more easily now!

That said, I'm glad this implementation no longer copy-pastes BehaviorSubject into this library. That's a complex class that handles a lot of different cases and has undergone a lot of small but important changes to make sure it works correctly.

I decided to take a shot at this myself without looking too much at what y'all had done, you can see my version here:  https://github.com/felangel/bloc/pull/839

Overall, the tests correctly catch a lot of the behavior (emitting first value, correctly handling errors, etc), so it feels pretty safe to make the change

packages/bloc/lib/src/bloc.dart
```diff
@@ -10,23 +9,24 @@ import '../bloc.dart';
 /// and transforms them into a `Stream` of `States` as output.
 /// {@endtemplate}
 abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
-  final PublishSubject<Event> _eventSubject = PublishSubject<Event>();
+  final _eventController = StreamController<Event>();
```
brianegan: Should be a broadcast Stream (PublishSubject is broadcast) -- this is why you had to change the test to check for a different error message, but that was correctly telling you something!

packages/bloc/test/bloc_test.dart
```diff
@@ -551,7 +551,7 @@ void main() {
           );
           expect(
             (capturedError as StateError).message,
-            'Cannot add new events after calling close',
+            'Cannot add event after closing',
```
brianegan: This test should not be changed, but a broadcast `StreamController` should be used which would give you the correct error.

mmozhaiskyi:
> Also, you can create a "Starts with Stream" really easily using `async*` functions so ya shouldn't need to copy the whole StartsWithStreamTransformer (another case where I really wouldn't copy-paste)

I am not sure. Approach which uses stream generator (`async*`) does not work correctly. I tried it before.
Please take a look to this tests. Values `1`, `2`, `3` never emitted when using stream generator.

```
import 'dart:async';

void main() {
  group('Start with test', () {
    test('start with using stream generator does not work', () {
      final controller = StreamController<int>.broadcast();
      final stream = _startWith(controller.stream, 0);

      expectLater(stream, emitsInOrder([0]));

      controller.add(1);
      controller.add(2);
      controller.add(3);

      controller.close();
    });

    test('start with using transformer generator works correctly', () {
      final controller = StreamController<int>.broadcast();
      final stream = controller.stream
          .transform(_StartWithStreamTransformer(0));

      expectLater(stream, emitsInOrder([0, 1, 2, 3]));

      controller.add(1);
      controller.add(2);
      controller.add(3);

      controller.close();
    });
  });
}

Stream _startWith(source, initialValue) async* {
  yield initialValue;
  yield* source;
}
```

packages/bloc/lib/src/bloc.dart
mmozhaiskyi: Ok, agreed. Many listeners inside `transformEvents` looks strange but it is possible. I will change it to broadcast stream. Thanks

packages/bloc/test/bloc_test.dart
mmozhaiskyi: Will do

brianegan:
Tis a fair point -- if you synchronously listen then start adding events right away, that will not work with a broadcast Stream. An alternative: Since the `State` stream is not being listened to more than once (folks are listening to the created Stream), you could use a non-broadcast StreamController for the State stream and it would function correctly. Like so:

```
import 'dart:async';

void main() {
  group('Start with test', () {
    test('start with using stream generator does not work', () {
      final controller = StreamController<int>();
      final stream = _startWith(controller.stream, 0);

      expectLater(stream, emitsInOrder([0, 1, 2, 3]));

      controller.add(1);
      controller.add(2);
      controller.add(3);

      controller.close();
    });
  });
}

Stream _startWith(source, initialValue) async* {
  yield initialValue;
  yield* source;
}
```

mmozhaiskyi:
> Tis a fair point -- if you synchronously listen then start adding events right away, that will not work with a broadcast Stream.

Hmm. Interesting 🤔. I don’t know it.
But what about [you proposal](https://github.com/felangel/bloc/pull/839)? You use broadcast stream with generator approach.

brianegan:
Yah, I didn't consider that exact use case -- some tests do something similar but passed fine with that current technique. Changed it up.

packages/bloc/lib/src/bloc.dart
```diff
@@ -187,3 +190,39 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
     onError(error, stacktrace);
   }
 }
+
+class _StartWithStreamTransformer<T> extends StreamTransformerBase<T, T> {
```
felangel: Can we just use an async generator here as Brian mentioned? It would greatly simplify this.

mmozhaiskyi:
Let`s make small conclusion:
1. `_stateController` should be broadcast stream to allow multi listeners for one bloc.
2. `_eventController` should be broadcast too to avoid problems inside custom implementation of `transformEvents` method (many subscriptions may listen to the stream of events inside a method).
3. Initial emit which contains `_currentState` is required now. But additional discussion would be good.
4. Only `_StartWithStreamTransformer` approach works correctly with broadcast streams. Unfortunately,  `async*` does not work. [See comment](https://github.com/felangel/bloc/pull/821#issuecomment-581130635)

I think latest version of code is simplified as it possible.

packages/bloc/lib/src/bloc.dart
mmozhaiskyi: Yes, we can. I will do. What we should doing with `Bloc.isBroadcast`. After changing `_stateController` to broadcast it always return `true`? But actually bloc still support many listeners like a broadcast stream.

mmozhaiskyi: We can return `true` always but it will be hardcoded value. Hmm. What do you think?

felangel: yeah we can just hardcode true since we can just enforce that within the bloc.

packages/bloc/lib/src/bloc.dart
```diff
@@ -10,23 +9,24 @@ import '../bloc.dart';
 /// and transforms them into a `Stream` of `States` as output.
 /// {@endtemplate}
 abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
-  final PublishSubject<Event> _eventSubject = PublishSubject<Event>();
+  final _eventController = StreamController<Event>.broadcast();
+  final _stateController = StreamController<State>();
 
-  BehaviorSubject<State> _stateSubject;
+  State _currentState;
```
felangel: nit but for consistency can we rename to `_state`?
```suggestion
  State _state;
```

packages/bloc/lib/src/bloc.dart
```diff
@@ -174,7 +179,8 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
         try {
           BlocSupervisor.delegate.onTransition(this, transition);
           onTransition(transition);
-          _stateSubject.add(nextState);
+          _stateController.add(nextState);
+          _currentState = nextState;
```
felangel: can we also set the state before adding to the state controller?

packages/bloc/lib/src/bloc.dart
```diff
@@ -10,23 +9,24 @@ import '../bloc.dart';
...
-  final PublishSubject<Event> _eventSubject = PublishSubject<Event>();
+  final _eventController = StreamController<Event>.broadcast();
+  final _stateController = StreamController<State>();
 
-  BehaviorSubject<State> _stateSubject;
+  State _state;
 
   /// Returns the current [state] of the [bloc].
-  State get state => _stateSubject.value;
+  State get state => _currentState;
```
felangel: this should be `_state` now too

packages/bloc/lib/src/bloc.dart
```diff
@@ -10,23 +9,24 @@ import '../bloc.dart';
...
 
   /// Returns whether the `Stream<State>` is a broadcast stream.
   @override
-  bool get isBroadcast => _stateSubject.isBroadcast;
+  bool get isBroadcast => true;
 
   /// {@macro bloc}
   Bloc() {
-    _stateSubject = BehaviorSubject<State>.seeded(initialState);
+    _currentState = initialState;
```
felangel: same here

packages/bloc/lib/src/bloc.dart
mmozhaiskyi: Done. Thank you for suggestion

packages/bloc/lib/src/bloc.dart
mmozhaiskyi: I am not sure what it can affect something, but was changed :)

packages/bloc/lib/src/bloc.dart
```diff
@@ -157,15 +157,20 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
   /// ```
   Stream<State> transformStates(Stream<State> states) => states;
 
+  Stream<State> _buildStateStream() async* {
+    yield _state;
+    yield* _stateController.stream;
+  }
+
   void _bindStateSubject() {
```
felangel: can we rename this `_bindEventsToStates` since we're no longer using subjects?

packages/bloc/lib/src/bloc.dart
```diff
@@ -10,23 +9,24 @@ import '../bloc.dart';
...
   /// Returns whether the `Stream<State>` is a broadcast stream.
   @override
-  bool get isBroadcast => _stateSubject.isBroadcast;
+  bool get isBroadcast => true;
 
   /// {@macro bloc}
   Bloc() {
-    _stateSubject = BehaviorSubject<State>.seeded(initialState);
+    _state = initialState;
     _bindStateSubject();
```
felangel:
```suggestion
    _bindEventsToStates();
```

packages/bloc/lib/src/bloc.dart
felangel:
```suggestion
  void _bindEventsToStates() {
```

mmozhaiskyi: Done

packages/bloc/lib/src/bloc.dart
```diff
@@ -77,7 +77,7 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
     try {
       BlocSupervisor.delegate.onEvent(this, event);
       onEvent(event);
-      _eventSubject.sink.add(event);
+      _eventController.sink.add(event);
```
felangel:
```suggestion
      _eventController.add(event);
```

packages/bloc/lib/src/bloc.dart
```diff
@@ -157,15 +157,20 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
   /// ```
   Stream<State> transformStates(Stream<State> states) => states;
 
-  void _bindStateSubject() {
+  Stream<State> _buildStateStream() async* {
+    yield _state;
+    yield* _stateController.stream;
+  }
```
tenhobi: This is cool! Very elegant. 💃

brianegan:
Wait, my bad -- was coding too quickly. My implementation has a big problem: You can't listen to the `state` stream twice! This test would fail with this current implementation. I would go back to using a `StreamController<State>.broadcast()` for the `stateStream`. It's also probably a good idea to add a test case that verifies the Stream can be listened to more than once.

```dart
      test('close does not emit new states over the state stream', () {
        final expectedStates = [equals(''), emitsDone];

        expectLater(
          simpleBloc,
          emitsInOrder(expectedStates),
        );

        expectLater(
          simpleBloc,
          emitsInOrder(expectedStates),
        );

        simpleBloc.close();
      });
```

This gets a bit hairy: I changed that state StreamController from a `broadcast` to a normal StreamController to prevent any errors using an `async*` for the `startWith` implementation, which were displayed by @mmozhaiskyi in his test code above. Of course, that means you can no longer listen to the Stream more than once 🤦‍♂ 

The cause of that issue in the tests? Events were added to the `StreamController` before the `async*` function could start listening (as the function is run asynchronously).

The `Bloc` class will not have that same problem, because events are not synchronously added to the `stateController`, but rather added asynchronously after being transformed by `transformEvent` and `transformState`. Therefore, the `async*` function has time to execute and start listening to the broadcast state Stream before any events would be sent. That's why this test actually works fine using the `async*` function for `startsWith` + a `broadcast` Stream on the Bloc class:

```dart
      test('should map single event to correct state', () {
        final expectedStates = ['', 'data'];

        expectLater(
          simpleBloc,
          emitsInOrder(expectedStates),
        );

        expectLater(
          simpleBloc,
          emitsInOrder(expectedStates),
        );

        expectLater(
          simpleBloc,
          emitsInOrder(expectedStates),
        );

        simpleBloc.add('event');
      });
```

brianegan:
Updated my branch with those changes!

packages/bloc/lib/src/bloc.dart
```diff
@@ -10,7 +10,7 @@ import '../bloc.dart';
 /// {@endtemplate}
 abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
   final _eventController = StreamController<Event>.broadcast();
-  final _stateController = StreamController<State>();
+  final _stateController = StreamController<State>.broadcast();
```
brianegan: Looks good! I'd just make sure to add the test to catch this bug as well.

felangel:
Going to merge and will apply the testing updates right after. Thanks everyone!

sla-000:
Hi @felangel So if I understand it right, though BehaviorSubject was removed from bloc/cubit but some other mechanism was implemented to keep it work as it was with BehaviorSubject?
I added listener to the cubit example from repo (see below) and I can see that listener does not get last value from cubit right after it connects to states stream. Listener only gets a new state if I call increment() after connection.
Is it supposed to be so or I do something wrong?

```
Future<void> cubitMain() async {
  print('----------CUBIT----------');

  /// Create a `CounterCubit` instance.
  final cubit = CounterCubit();

  /// Access the state of the `cubit` via `state`.
  print(cubit.state); // 0

  /// Interact with the `cubit` to trigger `state` changes.
  cubit.increment();

  await Future<void>.delayed(Duration.zero);

  final _sub = cubit.stream.listen((state) {
    print('Look, the state=$state');
  });

  await Future<void>.delayed(Duration.zero);

  cubit.increment();

  await Future<void>.delayed(Duration.zero);

  /// Access the new `state`.
  print(cubit.state);

  /// Close the `cubit` when it is no longer needed.
  await cubit.close();
  await _sub.cancel();
}
```

Log:
```
----------CUBIT----------
onCreate -- bloc: CounterCubit
0
onChange -- bloc: CounterCubit, change: Change { currentState: 0, nextState: 1 }
onChange -- bloc: CounterCubit, change: Change { currentState: 1, nextState: 2 }
Look, the state=2
2
onClose -- bloc: CounterCubit
```

felangel:
Hi @sla-000 this is working as expected and has been this way since bloc v6.0.0. See https://bloclibrary.dev/#/migration?id=%e2%9d%97bloc-does-not-emit-last-state-on-subscription for more information.

sla-000:
Hi @felangel Thanks for quick response! 
So from v6.0.0 if I want to subscribe a bloc to changes in other bloc I have to pass both current state value and a stream to this bloc and then 
1 - process state, if there is one, it's also must be checked
2- subscribe to stream and process new data in a listener 

This creates boilerplate code. Is there any chance this ever will be changed back to BehaviorSubject or similar?  I understand you are trying to keep code pure with as many dependencies as possible but may be it is too much 🙂

maximveksler:
@sla-000 have you considered using a shared repository to implement communication between the blocks? This way it might simplify your use case of needing to cope with the duality of the bloc behaviour.

Note that `BlocBuilder` and `BlocListener` will perform correctly, for both 1st event and followup events.

sla-000:
@maximveksler Could you provide some example of that shared repository, please?

maximveksler:
Yes, however I can't testify that it's the canonical example. 

I will reference you the 2 examples, 1st is the firebase login where the authentication repository is shared between the App, Login and Signup BloC's. Pay attention that the App BloC and the Login Cubit listen to the user stream from the login repository.

- User stream https://github.com/felangel/bloc/blob/master/examples/flutter_firebase_login/packages/authentication_repository/lib/src/authentication_repository.dart#L183
- App BloC listen https://github.com/felangel/bloc/blob/master/examples/flutter_firebase_login/lib/app/bloc/app_bloc.dart#L22
- Login BloC communication https://github.com/felangel/bloc/blob/997cfa8835d5e2378bb770f45d970f457cf664fa/examples/flutter_login/lib/login/bloc/login_bloc.dart#L55


The 2nd would be the Todo's example, where 2 BloC's are listening for todo's https://github.com/felangel/bloc/blob/master/examples/flutter_todos/lib/stats/bloc/stats_bloc.dart#L24 & https://github.com/felangel/bloc/blob/master/examples/flutter_todos/lib/todos_overview/bloc/todos_overview_bloc.dart#L31 while another bloc is adding new ones https://github.com/felangel/bloc/blob/master/examples/flutter_todos/lib/edit_todo/bloc/edit_todo_bloc.dart#L52

PR #827, opened 2020-01-31
Main Ci improvements

tenhobi:
macOS does some weird things in the setup and this change only runs the flutter action when there have been some changes, otherwise, finish

Also, the caching on macOS seems very strange and it still does some work for some reason and is slower than Ubuntu.

image

I would consider disabling macOS testing, at least for now. I have also seen some warning about the size of the cache, as we can only use 2 GB for now, so keeping only ubuntu cache would be better...

Also, we might want to remove `setup` job at all? Now we have to manually change the used flutter version, which means only that CI will be slower (installing the new version) and other commits will have the cached version all the time. We have `setup` now to eliminate the need of downloading new version in each matrix job, but that adds 1 m of time to each CI check. And I think we dont care about some warnings that there is already a cache with this ID etc. once we increase version number... What do you think @felangel ?


tenhobi:
So, after some research, the increased time in macOS is because it cannot store new cache as the size combined with ubuntu is more than 2GB. 

With all the changes (at least for now), removing setup job and macOS, it runs "very fast":

image

We can add macOS later, when GitHub Actions are more evolved and provide more caching space :-)

.github/workflows/main.yaml
```diff
@@ -7,50 +7,19 @@ env:
   flutter_version: '1.12.13+hotfix.7'
```
felangel: Do we still need flutter version?

.github/workflows/main.yaml
```diff
@@ -7,50 +7,19 @@ env:
...
-
   # Analyze and test each package.
   analyze-and-test:
     runs-on: ${{ matrix.os }}
-    needs: setup
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest] # macos-latest
```
felangel: Can we just delete macos for now?

.github/workflows/main.yaml
tenhobi: Yes.

.github/workflows/main.yaml
tenhobi: Sure.

PR #838, opened 2020-02-02
Decouple Bloc from RxDart

felangel:
## Status
**IN DEVELOPMENT**

## Breaking Changes
YES

## Description
- Remove `rxdart` as a dependency from `bloc` (related to #821)
- Refactor to use `_TransitionStreamTransformer` object which encapsulates the current transition in order to ensure that transitions occur for the respective event/state pair.(https://github.com/felangel/bloc/pull/839#discussion_r373839896)
- Rename `transformStates` to `transformTransitions`
- `transformEvents` signature change to take a `TransitionFunction`

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

- Breaking changes
  - `transformStates` -> `transformTransitions`
  - Function signature changes to `transformEvents`

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -4,6 +4,11 @@ import 'package:meta/meta.dart';
 
 import '../bloc.dart';
 
+/// Signature for a mapper function which takes an [Event] as input
+/// and outputs a [Stream] of [Transition] objects.
+typedef TransitionFunction<Event, State> = Stream<Transition<Event, State>>
+    Function(Event);
+
 /// {@template bloc}
 /// Takes a `Stream` of `Events` as input
 /// and transforms them into a `Stream` of `States` as output.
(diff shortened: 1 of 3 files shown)


felangel:
@brianegan when you have a min can you take a look?

packages/bloc/lib/src/bloc.dart
```diff
@@ -142,39 +154,42 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
...
+    Stream<Transition<Event, State>> transitions,
+  ) {
+    return transitions;
+  }
+
+  void _bindEventsToStates() {
+    transformTransitions(
+      transformEvents(_eventController.stream, (event) {
+        return mapEventToState(event)
+            .transform(_TransitionStreamTransformer(this, event))
```
mmozhaiskyi: Can you just use `map` operator in this case? `_TransitionStreamTransformer` looks complicated.

```
.map((nextState) => 
  Transition(
    currentState: _state,
    event: event,
    nextState: nextState
  )
)
```

felangel: yup good point. I originally had it due to the `StartWith` changes but it's no longer needed 👍

PR #839, opened 2020-02-02
Remove RxDart dependency

brianegan:
## Status
**READY/IN DEVELOPMENT/HOLD**

## Breaking Changes
YES | NO

## Description
A few sentences describing the overall goals of the pull request's commits.

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------
other_pr_production | [link]()
other_pr_master | [link]()

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Examples

## Steps to Test or Reproduce

```
```

This PR will affect:

*

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -1,7 +1,6 @@
 import 'dart:async';
 
 import 'package:meta/meta.dart';
-import 'package:rxdart/rxdart.dart';
 
 import '../bloc.dart';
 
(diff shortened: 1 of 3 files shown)


packages/bloc/lib/src/bloc.dart
```diff
@@ -41,14 +43,20 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
...
+    return _createStateStream().listen(
       onData,
       onError: onError,
       onDone: onDone,
       cancelOnError: cancelOnError,
     );
   }
 
+  // Creates a Stream that emits the current state and any future states
+  Stream<State> _createStateStream() async* {
```
brianegan: One thing that came up for me: How important is it that blocs emit the latest event when you first listen? Is it even necessary?

I know that'd be a bigger breaking change to discuss, but seems like most blocs are hooked up to `StreamBuilder` widgets. Since you already have the `initialState` to pass to the `StreamBuilder` for the initial build, is it really necessary for this class to emit the latest value to all listeners (triggering 2 builds for each BlocBuilder widget?)

packages/bloc/lib/src/bloc.dart
```diff
@@ -157,15 +165,15 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
   /// ```
   Stream<State> transformStates(Stream<State> states) => states;
 
-  void _bindStateSubject() {
+  void _bindEventsToStates() {
     Event currentEvent;
```
brianegan: General comment: It feels like this could lead to bugs where `asyncExpand` is not used in `transformEvents`, and therefore events will be delivered to the `forEach` that might respond to different events.

felangel: Hmm yeah good point. Haven't run into that case so far but I see what you're saying.

packages/bloc/lib/src/bloc.dart
felangel: Haha I was thinking about that yesterday as well. I think keeping the behavior as is for now makes sense. I need to think through what the implications of removing that are because actually BlocBuilder and BlocListener already use skip(1) to avoid the 2 builds issue.

packages/bloc/lib/src/bloc.dart
felangel: thoughts on https://github.com/felangel/bloc/pull/838?

felangel:
Closing this since #821 was merged. Thanks so much for your help as always! 🙏

PR #840, opened 2020-02-02
Bloc Internal Mapping to Transition

felangel:
## Status
**READY**

## Breaking Changes
YES

## Description
- Refactor to bloc to map directly to `Transition` objects which encapsulate the current transition in order to ensure that transitions occur for the respective event/state pair -- addresses https://github.com/felangel/bloc/pull/839#discussion_r373839896.
- Rename `transformStates` to `transformTransitions`
- `transformEvents` signature change to take a `TransitionFunction`

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

- Breaking changes
  - `transformStates` -> `transformTransitions`
  - Function signature changes to `transformEvents`


felangel:
@brianegan when you have a min can you please take a look and lmk what you think? Thanks!

packages/bloc/lib/src/bloc.dart
```diff
@@ -141,51 +153,53 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
...
-        );
-        try {
-          BlocSupervisor.delegate.onTransition(this, transition);
-          onTransition(transition);
-          _state = nextState;
-          _stateController.add(nextState);
-        } on dynamic catch (error) {
-          _handleError(error);
-        }
+    transformTransitions(transformEvents(
```
brianegan: Nice!

packages/bloc/lib/src/bloc.dart
```diff
@@ -141,51 +153,53 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
...
+            currentState: state,
+            event: event,
+            nextState: nextState,
+          );
+        }).skipWhile((transition) {
+          return state == transition.nextState || _stateController.isClosed;
+        }).handleError(_handleError);
       },
-    );
+    )).forEach((transition) {
```
brianegan: One other random Q: instead of using `forEach`, should this just `listen` to the Stream and store a `StreamSubscription _transitionSubscription` for cancellation in the `close` method?

felangel: yeah good call 👍

PR #851, opened 2020-02-07
Build iOS app + other CI enhancements

### LINKED ISSUE #835: Enable CI for Mac & iOS
In #827, the github actions were disabled for Mac as they were slow and not very stable.
We couldn't cannot cache flutter on both systems (2 GB limit).

This FR is to reevaluate running the CI on Mac and building iOS applications for the examples, once (hopefully) Github Actions provide more caching.

cc @tenhobi

tenhobi:
## Status
**IN DEVELOPMENT**

## Breaking Changes
NO

## Description

In v1.1.1 of actions/cache https://github.com/actions/cache/releases/tag/v1.1.1 there has been an update to 5 GB of cache space (!!!), so now we can fully use the CI and do some things also in macos.

I think keeping testing and analyzing only in one OS would be sufficient, so we will only add a macos support to Flutter build task.

This PR does:

- [x] slightly updated documentation and naming
- [x] fix wrong uploaded coverage for examples (`${{ matrix.package }}` to `${{ matrix.folder }}`)
- [x] add `ios` build option to CI script
- [x] create `examples-pure` job to cover pure Dart examples (as is bloc/example and bloc_test/example)
- ~adds an example google services file for iOS~
- [x] temporarily removed all builds due performance reasons

Fixes #835 

~I would need a hand from @felangel or @netantho to create the last step and providing an example google services file for iOS and make it work. :-)~

**Note that the diff is unfortunatelly a bit messy, due to reordering** so we can have in order: packages, examples-pure, examples-web and at least examples-flutter.


.github/workflows/main.yaml
```diff
@@ -57,15 +58,15 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          file: ${{ matrix.package }}.lcov
+          file: ${{ matrix.folder }}/coverage/lcov.info
```
felangel: 💯

tenhobi:
Note: Due to performance issues running CI, we (temporarily) disabling building flutter & web.

PR #894, opened 2020-02-18
update README.md

tohami:
change provider create to be builder

## Status
**READY**

## Breaking Changes
NO

## Description
update README.md samples to match the current code state by replace create: (context) to be builder: (context)

This PR will affect: just README.md file
*


felangel:
Hi @tohami 👋
Where did you see builder being used in favor of create?
Create replaced builder in order to maintain parity with provider so whichever sample you were referencing is outdated and we should update the sample rather than the docs.

PR #906, opened 2020-02-20
Clean up and improve extension snippets

zepfietje:
## Status
**READY**

## Breaking Changes
NO

## Description
This PR cleans up the extension snippets to reflect the latest changes in the Bloc packages (e.g. naming conventions and extension methods). The `blocstate` and `blocevent` snippets have been removed in favor of the already present templates. Possibly, I'll open another PR with some more snippets for frequently used states and events (like in the naming convention docs).

On another note, the docs on snippets seem to be very outdated. Shall we update or remove them? What do you suggest?


extensions/vscode/snippets/dart.json
```diff
@@ -0,0 +1,166 @@
+{
```
felangel: why did we rename this file to `dart.json`? We will need to update the `package.json` to point to the new snippets file. I would prefer to leave the file name as it was unless there's a specific reason for the rename.

zepfietje: Previously (when the file was named `snippets.json`), all snippets needed to specify a scope (`"scope": "dart"`). This is no longer required when renaming the file to `dart.json`.

Now I see `"language": "dart"` is also specified in the `package.json`. Although, I'm not sure if this works the same way when keeping `snippets.json` and removing the scopes from the individual snippets.

zepfietje:
> Also, might be good to remove most of the gifs from the README since it's hard to keep them up-to-date.

Maybe we could just remove all mentions of specific snippets and link to the available snippets (in both the README and docs)?

felangel:
@zepfietje can we do the file rename in a separate PR if it's necessary? It's harder to review because the diff isn't clean. I'm also fine with having a single source of truth for all supported snippets and linking 👍

zepfietje:
@felangel, I've split the changes into separate commits for easier review.

PR #913, opened 2020-02-24
bloc_test v4.0.0

### LINKED ISSUE #910: Make blocTest less verbose and usable for more occasions
> ### Warning ⚠
>None of what you see below may be applicable to a reasonable Flutter developer. It may well be the case that I'm the only one who needs the following changes. However, I plan to use the changed `blocTest` in some of my tutorials.

First up, I find that having to specify initial states in every `blocTest` is redundant. It would be better if they were skipped. Testing the initial state is always possible with the `initialState` property in a regular `test`, no need to create a `blocTest`.

Secondly, it's useful to have the `Bloc` instance readily available inside the `verify` function. Consider:
```dart
blocTest(
  "should register using IAuthFacade if email and password are valid",
  build: () async {
    arrangeAuthFacadeSuccess();
    await prefillWithValidEmailAndPassword();
    return signInFormBloc;
  },
  act: (bloc) async =>
      bloc.add(const SignInFormEvent.registerWithEmailAndPasswordPressed()),
  verify: (SignInFormBloc bloc) async {
    verify(mockAuthFacade.registerWithEmailAndPassword(
      emailAddress: argThat(
        equals(bloc.state.validatedEmailAddress),
        named: 'emailAddress',
      ),
      password: argThat(
        equals(bloc.state.validatedPassword),
        named: 'password',
      ),
    ));
  },
);
```

The above snippet of code brings me to my third point - making the `build` function asynchronous. With Blocs that have only one state class, such as the one from the [Firebase login tutorial](https://bloclibrary.dev/#/flutterfirebaselogintutorial), it may be useful to initialize (or "prefill") the Bloc into a certain state. See the prefilling method:

```dart
  Future<void> prefillWithValidEmailAndPassword() async {
    signInFormBloc.add(const SignInFormEvent.emailChanged('abc@gmail.com'));
    signInFormBloc.add(const SignInFormEvent.passwordChanged('pazzwrd'));
    await signInFormBloc.take(3).last;
  }
```

As the events which "prefill" the Bloc will trigger state changes in which I'm not interested as they're tested separately, I need to ignore them using the `take(number).last` contraption. This is the reason why the `build` function has to be `async`.

Here's my proposed modified `blocTest`. I'm open to splitting it up into a separate function as not to force everyone into using it.

```dart
@isTest
void blocTest<B extends Bloc<Event, State>, Event, State>(
  String description, {
  @required Future<B> Function() build,
  // not required - sometimes we want to just verify without expecting
  Iterable expect,
  Future<void> Function(B bloc) act,
  Duration wait,
  Future<void> Function(B bloc) verify,
}) {
  test.test(description, () async {
    final bloc = await build();
    final states = <State>[];
    // Skipping the initial state
    final subscription = bloc.skip(1).listen(states.add);
    await act?.call(bloc);
    if (wait != null) await Future.delayed(wait);
    await bloc.close();
    if (expect != null) test.expect(states, expect);
    await subscription.cancel();
    await verify?.call(bloc);
  });
```

felangel:
## Status
**READY**

## Breaking Changes
YES

## Description
Closes #910

#### Async Build

**Before:**
```dart
blocTest(
  '...',
  build: () => MyBloc(),
  ...
);
```

**After:**
```dart
blocTest(
  '...',
  build: () async => MyBloc(),
  ...
);
```

#### Skip InitialState by Default

**Before:**
```dart
blocTest(
  build: () async => CounterBloc(),
  act: (bloc) => bloc.add(CounterEvent.increment),
  expect: [0, 1],
);
```

**After:**
```dart
blocTest(
  build: () async => CounterBloc(),
  act: (bloc) => bloc.add(CounterEvent.increment),
  expect: [1],
);
```

#### Optional Skip (defaults to 1)
```dart
blocTest(
  build: () async => CounterBloc(),
  act: (bloc) => bloc.add(CounterEvent.increment),
  skip: 0,
  expect: [0, 1],
);
```

#### Verify Accepts Bloc as Parameter

**Before:**
```dart
blocTest(
  ...
  verify: () async { ... }
);
```

**After:**
```dart
blocTest(
  ...
  verify: (bloc) async { ... }
);
```

#### Skip InitialState by Default

**Before:**
```dart
test('emits [0, 1] when CounterEvent.increment is added', () async {
  final bloc = CounterBloc();
  bloc.add(CounterEvent.increment);
  await emitsExactly(bloc, [0, 1]);
});
```

**After:**
```dart
test('emits [1] when CounterEvent.increment is added', () async {
  final bloc = CounterBloc();
  bloc.add(CounterEvent.increment);
  await emitsExactly(bloc, [1]);
});
```

#### Optional Skip (defaults to 1)
```dart
test('emits [0, 1] when CounterEvent.increment is added', () async {
  final bloc = CounterBloc();
  bloc.add(CounterEvent.increment);
  await emitsExactly(bloc, [0, 1], skip: 0);
});
```

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

- Breaking changes to `blocTest` which will be included in v4.0.0.

### DIFF
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -121,7 +121,7 @@ void blocTest<B extends Bloc<Event, State>, Event, State>(
     await act?.call(bloc);
     if (wait != null) await Future.delayed(wait);
     await bloc.close();
-    test.expect(states, expect);
+    if (expect != null) test.expect(states, expect);
     await subscription.cancel();
     await verify?.call(bloc);
   });
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -11,10 +11,14 @@ import 'package:test/test.dart' as test;
 /// by closing the [bloc] stream before evaluating the [expect]ation.
 ///
 /// [build] should be used for all [bloc] initialization and preparation
-/// and must return the [bloc] under test.
+/// and must return the [bloc] under test as a `Future`.
 ///
 /// [act] is an optional callback which will be invoked with the [bloc] under
 /// test and should be used to `add` events to the [bloc].
+///
+/// [skip] is an optional `int` which can be used to skip any number of states.
+/// The default value is 1 which skips the `initialState` of the bloc.
+/// [skip] can be overridden to include the `initialState` by setting it to 0.
 ///
 /// [wait] is an optional `Duration` which can be used to wait for
 /// async operations within the [bloc] under test such as `debounceTime`.
@@ -23,14 +27,15 @@ import 'package:test/test.dart' as test;
 /// under test is expected to emit after [act] is executed.
 ///
 /// [verify] is an optional callback which is invoked after [expect]
-/// and can be used for additional non-bloc related assertions.
+/// and can be used for additional verification/assertions.
+/// [verify] is called with the [bloc] returned by [build].
 ///
 /// ```dart
 /// blocTest(
-///   'CounterBloc emits [0, 1] when CounterEvent.increment is added',
-///   build: () => CounterBloc(),
+///   'CounterBloc emits [1] when CounterEvent.increment is added',
+///   build: () async => CounterBloc(),
 ///   act: (bloc) => bloc.add(CounterEvent.increment),
-///   expect: [0, 1],
+///   expect: [1],
 /// );
 /// ```
 ///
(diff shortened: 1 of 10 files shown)


packages/bloc_test/lib/src/bloc_test.dart
```diff
@@ -78,30 +98,31 @@ import 'package:test/test.dart' as test;
...
   test.test(description, () async {
-    final bloc = build();
+    final bloc = await build();
     final states = <State>[];
-    final subscription = bloc.listen(states.add);
+    final subscription = bloc.skip(skip).listen(states.add);
     await act?.call(bloc);
     if (wait != null) await Future.delayed(wait);
     await bloc.close();
     test.expect(states, expect);
```
ResoDev: Missing null check for the no-longer `@required` `expect` parameter.

felangel: good catch 👍

PR #965, opened 2020-03-15
Add optional dispose callback to RepositoryProvider

### LINKED ISSUE #952: [Best practice] Release resource with RepositoryProvider
Hi, @felangel 

I release resources in the bloc which is provided by BlocProvider.

```
@override
  Future<void> close() async {
    super.close();
  }
```

What's the best practice to release resources in the repository which is provided by RepositoryProvider?

Thanks

felangel:
## Status
**READY**

## Breaking Changes
NO

## Description
- Add optional `dispose` callback to `RepositoryProvider`
- Closes #952 

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

New feature which will be included in `flutter_bloc v3.3.0`

### DIFF
--- a/docs/flutterbloccoreconcepts.md
+++ b/docs/flutterbloccoreconcepts.md
@@ -256,6 +256,18 @@ context.repository<RepositoryA>();
 RepositoryProvider.of<RepositoryA>(context)
 ```
 
+The repository can be optionally disposed via the `dispose` callback. `dispose` will be called when `RepositoryProvider` is unmounted from the widget tree.
+
+```dart
+RepositoryProvider(
+  create: (context) => RepositoryA(),
+  dispose: (context, repository) {
+    // dispose logic goes here
+  }
+  child: ChildA(),
+);
+```
+
 ### MultiRepositoryProvider
 
 **MultiRepositoryProvider** is a Flutter widget that merges multiple `RepositoryProvider` widgets into one.
--- a/packages/flutter_bloc/CHANGELOG.md
+++ b/packages/flutter_bloc/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 3.3.0
+
+- Add optional `dispose` callback to `RepositoryProvider` ([#952](https://github.com/felangel/bloc/issues/952))
+
 # 3.2.0
 
 - Fix type inference for: `MultiBlocProvider`, `MultiRepositoryProvider`, `MultiBlocListener` ([#773](https://github.com/felangel/bloc/pull/773))
--- a/packages/flutter_bloc/README.md
+++ b/packages/flutter_bloc/README.md
@@ -257,6 +257,18 @@ context.repository<RepositoryA>();
 RepositoryProvider.of<RepositoryA>(context)
 ```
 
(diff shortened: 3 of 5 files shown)


jorgecoca:
What would be the use case to `dispose` a Repository? It was my understanding that these components are not tied to a lifecycle, and ideally, they should not have subscriptions on their own... I wonder if these changes might introduce/hide a code smell, rather than providing good, useful functionality.

RollyPeres:
Since repositories should just interact with different data sources, they could indeed receive some dependencies but it's not their responsibility to get rid of those(if needed), but the layer creating them.
Repositories should also not contain any subscriptions as very well pointed out by @jorgecoca , but instead plain old `Stream`s; subscriptions should happen inside blocs, or alternatively through a `StreamProvider`.
I personally don't think they need a `dispose` method, since that would imply that repositories have lifecycle/some sort of state.

felangel:
@jorgecoca, @RollyPeres good callouts 👍 
Waiting on #952 to get more details around the use-case but will likely close this.

alex-isaienko-uapay:
I'm very sad to see this PR closed.

My original (maybe somewhat wrong) idea idea of the repository is was that it is an "abstraction layer over data access that provides a meaningful and easy-to-use API for consuming the business models".

I have two cases where not having a `dispose` method in `RepositoryProvider` hurts the app architecture:

### Using the [Floor DB](https://pub.dev/packages/floor)
I need a place where I can call "db.close()".

I'm quite sure that this thing is even more important for use-cases that involve constant connection to some data source (WebSockets, for example).

### Second (RAM) layer of cache
To ensure that the data is available to the user as fast as possible, I pre-cache a lot of stuff from DB cache into RAM cache at various points of the app lifecycle. The RAM cache is an observable data source itself, so it also needs to be closed at some point.

### Current workarounds
Right now, I see three options how to overcome the problem:
1. Provide the database in the tree above RepositoryProvider, using vanilla Provider. It would work, but this way the children in the tree get access to the data source directly. Not good, encapsulation suffers, data access details leak to the children widgets.
2. Have the data source as a private singleton defined within the same file as a repository. Not good, no good way of replacing the data source for testing, won't work in cases where some supervision over the data source is needed (like maintaining the WebSocket connected).
3. Ignore the problem and have RepositoryProvider at the top of the widget tree. Not good, might need to provide the same repository type in a more local scope in the future.

@felangel what do you think about my use-case, maybe you know a better way of achieving this without the `dispose` callback?

pmartelletti:
Hi @felangel ! I'm confused about why this was discarded - maybe best practices changed? On one hand, you say that Repositories should not have any state nor be in charge of disposing/closing any connection/subscription.

However, on this talk you gave at Flutter Forward 2023, you give an example that does exactly that (https://www.youtube.com/watch?v=TGKipiJhpXo&ab_channel=FlutterCommunity). The WorldRepository holds a subscription to a WebSocket which is closed when the `close` method is called on the repository.

So which one is it? I'm genuinely asking as I'm confused right now. I do like the idea of keeping the repository clean, but the idea of having a bloc just to create the WS connection and closing it on dispose bothers me a bit (there's not much logic in that bloc, if you ask me), so any guidance will be more than appreciated. 😄 

Thanks!

PR #1011, opened 2020-04-07
Remove try catch block in Auth Bloc in Firebase Login example

ElasticBottle:
I noticed that in the for state mapping for AppStarted event, the Try Catch block wraps the  _userRepository.isSignedIn(); 

This seems unnecessary since that repository just calls firebase.instance.currentUser() under the hood. The method only returns a Firebase User or a null value.

## Status
**READY**

## Breaking Changes
NO

## Description
Remove unnecessary try catch block in firebase login example

## Todos
Nill

## Steps to Test or Reproduce
Nill

This PR will affect:

* simplify firebase login example

### DIFF
--- a/examples/flutter_firebase_login/lib/authentication_bloc/authentication_bloc.dart
+++ b/examples/flutter_firebase_login/lib/authentication_bloc/authentication_bloc.dart
@@ -32,7 +32,6 @@ class AuthenticationBloc
   }
 
   Stream<AuthenticationState> _mapAppStartedToState() async* {
- 
       final isSignedIn = await _userRepository.isSignedIn();
       if (isSignedIn) {
         final name = await _userRepository.getUser();
@@ -40,7 +39,6 @@ class AuthenticationBloc
       } else {
         yield Unauthenticated();
       }
-    
   }
 
   Stream<AuthenticationState> _mapLoggedInToState() async* {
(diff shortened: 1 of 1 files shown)


examples/flutter_firebase_login/lib/authentication_bloc/authentication_bloc.dart
```diff
@@ -32,17 +32,15 @@ class AuthenticationBloc
   }
 
   Stream<AuthenticationState> _mapAppStartedToState() async* {
-    try {
+
```
felangel, line 35:
```suggestion
```

can we remove the extra newlines?

examples/flutter_firebase_login/lib/authentication_bloc/authentication_bloc.dart
```diff
@@ -32,17 +32,15 @@ class AuthenticationBloc
...
       if (isSignedIn) {
         final name = await _userRepository.getUser();
         yield Authenticated(name);
       } else {
         yield Unauthenticated();
       }
-    } catch (_) {
-      yield Unauthenticated();
-    }
+
```
felangel, line 43:
```suggestion
```

ElasticBottle:
All done!

PR #1017, opened 2020-04-10
package:bloc v4.0.0-dev.2

felangel:
## Status
**READY**

## Breaking Changes
YES

## Description
- Bug Fix: Do not emit duplicate terminating state
- **Breaking Change: Always emit the first state even if it is equal to the `initialState`**

## Todos
- [X] Tests
- [ ] Documentation
- [ ] Examples

- breaking changes in `package:bloc` which will potentially be included in v4.0.0-dev.2


orsenkucher:
@felangel 
Looks almost perfect to me🥰

Yet we should consider putting `.distinct(...` before `transformTransitions`, but after `transformEvents`.

For instance, if `transformTransitions` is throttling based on `nextState` for any reason, it will be knocked back every time by duplicate state.

And I really want to invent a test which will break if `.where((_) => !_stateController.isClosed)` gets removed.

packages/bloc/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 4.0.0-dev.2
+
+- Bugfix: Do not emit duplicate terminating state
+- Breaking Change: Always emit the first state even if it is equal to the `initialState`
```
jorgecoca, line 4: What would be the reasoning behind this change? Is there a use case for it?

orsenkucher: Please take a look over following snippet
```dart
BlocBinder<ValidationBloc, ValidationState, FeedBloc, FeedState>(
  direct: (context, state, bloc) { state. ...some stuff },
  reverse: (context, state, bloc) {
    state.when(
      available: (tree) => bloc.add(ValidationEvent.focus(tree.question)),
      empty: () {},
    );
  },
),
```
Some important business logic is done between blocs here. As its important business logic, seeded state will be always skipped to prevent unintended side-effects.
Then in `FeedBloc`'s constructor, we say 

> if I have cached tree available - rethrow it,
> otherwise ask questionsBloc for a new one

```dart
class FeedBloc extends HydratedBloc<FeedEvent, FeedState> {
final QuestionsBloc questionsBloc;
FeedBloc(...) {
  state.when(
    available: (tree) => add(FeedEvent.newArrived(tree)),
    empty: () => questionsBloc.add(QuestionsEvent.fetch()),
  );
}
...
```
And now we have arrived to the problem: `FeedBloc` won't emit, as it's already hydrated (seeded) with this `state`.
Nothing will reach our important business logic.

orsenkucher: About the reasoning, I will rephrase a bit
We are allowing **the every first state** to be emitted, despite what was in seeded state.
Shifting **distinct invariant** to be applied only to `mapEventToState` states, as business logic should not be affected by some discarded seeded values

PR #1019, opened 2020-04-10
Added Japanese translation except the example apps

dshukertjr:
## Status
**READY**

## Breaking Changes
NO

## Description
Added Japanese to all the files except the example apps. 

## Todos
- [ ] Tests
- [x] Documentation
- [ ] Examples


docs/jp/faqs.md
```diff
@@ -0,0 +1,258 @@
...
+2. 依存性は注入可能で、プラットフォームに依存しない。
+3. プラットフォームに分岐してはならない。
+4. 上記のルールを守っている限り実装方法はどのような形でも良い。
+
+UIのガイドラインは:
+
+1. 一つ一つの「それなりに複雑な」コンポーネントは Bloc を持つべき。
+2. コンポーネントは入力をありのまま送るべき。
+3. コンポーネントは出力を出来るだけありのまま表示するべき。
+4. 全ての分岐はシンプルに BloC の主力に関連するべき。
```
dshukertjr, line 155: I am not very confident about the translation here, because I feel like I don't fully understand the line here that says
"4. All branching should be based on simple BLoC boolean outputs.".

Could you elaborate on this and explain what exactly you mean by this sentence?

felangel: This means you should avoid having complex logic to determine what to render. Instead, the UI's logic should be really simple like

```dart
if (state is X) return WidgetA();
if (state is Y) return WidgetB();
```

Let me know if that helps 👍

dshukertjr: Thank you! That helps a lot. Let me change the Japanese sentence there in a minute.

felangel:
@dshukertjr thanks so much for kicking off these translations! Let me know when you're ready and I can review and merge them 👍

dshukertjr:
So with the last change, I think my translation is good to go.

I have not done the examples since there were a lot, and I was going to do them later probably in a separate pull request. Is that okay? Or do you want to wait for the example app to be translated as well before you merge them?

felangel:
> @felangel So with the last change, I think my translation is good to go.
> 
> I have not done the examples since there were a lot, and I was going to do them later probably in a separate pull request. Is that okay? Or do you want to wait for the example app to be translated as well before you merge them?

Awesome, will review shortly and I'm totally fine doing the examples in a separate PR 😄 
Thanks again, I really appreciate it 💯

PR #1034, opened 2020-04-12
Add mapState to bloc_test

lucasnlm:
## Status
READY

## Breaking Changes
NO

## Description
Makes `blocTest` simpler by adding a new `mapState` field to it.
That change allow simpler error messages like:

```dart
  Expected: [1, 2, 33]
    Actual: [1, 2, 3]
     Which: was <3> instead of <33> at location [2]
```

Where we are comparing only the wanted part of the state.

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------

## Todos
- [x] Tests
- [x] Documentation
- [x] Examples

## Steps to Test or Reproduce

- It adds `mapState` to `blocTest`.

```dart
      blocTest(
        'emits [1, 2, 3] when three ComplexEventA are added',
        build: () async => ComplexBloc(),
        act: (bloc) async {
          bloc
            ..add(ComplexEventA(1))
            ..add(ComplexEventA(2))
            ..add(ComplexEventA(3));
        },
        mapState: (state) => state.id,
        expect: [1, 2, 3],
      );
```

This PR will affect:

*  `blocTest` from `bloc_test`

### DIFF
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -26,6 +26,9 @@ import 'package:test/test.dart' as test;
 /// [expect] is an `Iterable` of matchers which the [bloc]
 /// under test is expected to emit after [act] is executed.
 ///
+/// [mapState] is an optional function used to transform the state
+/// in a simpler object. It may be used to make [expect] easier.
+///
 /// [verify] is an optional callback which is invoked after [expect]
 /// and can be used for additional verification/assertions.
 /// [verify] is called with the [bloc] returned by [build].
@@ -92,6 +95,20 @@ import 'package:test/test.dart' as test;
 /// );
 /// ```
 ///
+/// [blocTest] can also be used to transform the result states insimpler
+/// testable objects.
+/// 
+/// ```dart
+/// blocTest(
+///   'ComplexBloc emits [0, 1] when ComplexEvent() is added',
+///   build: () async => ComplexBloc(),
+///   act: (bloc) => bloc.add(ComplexEvent()),
+///   skip: 0,
+///   mapState: (state) => state.interestPart,
+///   expect: [0, 1],
+/// );
+/// ```
+///
 /// **Note:** when using [blocTest] with state classes which don't override
 /// `==` and `hashCode` you can provide an `Iterable` of matchers instead of
 /// explicit state instances.
@@ -112,6 +129,7 @@ void blocTest<B extends Bloc<Event, State>, Event, State>(
   Duration wait,
   int skip = 1,
   Iterable expect,
+  dynamic Function(State state) mapState,
   Future<void> Function(B bloc) verify,
 }) {
   test.test(description, () async {
@@ -121,7 +139,12 @@ void blocTest<B extends Bloc<Event, State>, Event, State>(
     await act?.call(bloc);
     if (wait != null) await Future.delayed(wait);
     await bloc.close();
-    if (expect != null) test.expect(states, expect);
+    if (expect != null) {
+      test.expect(
+        mapState != null ? states.map(mapState).toList() : states,
+        expect,
+      );
+    }
     await subscription.cancel();
     await verify?.call(bloc);
   });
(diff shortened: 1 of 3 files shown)


felangel:
Hi @lucasnlm 👋
Thanks for opening a PR! If the goal of this PR is to make the actual message easier to read, then I would recommend overriding toString on the state rather than adding an optional mapState to the blocTest. Let me know what you think!

lucasnlm:
No. It's not about `toString()`. Actually the objective it make the easier to compare the values in `expect` with the result of complex states. I know is possible make custom `Matcher` (like `isA<ComplexStateB>()`). But a `Matcher` is not soo simple.

felangel:
@lucasnlm why not use Equatable, freezed, or built value?

lucasnlm:
Using Equatable, I may get something like:

```
  Expected: [
              GameState:GameState(Status.running, 0, Item(2, 16, 1), ..., Vec2d(0, -1), Board(10, 20))
            ]
    Actual: [
              GameState:GameState(Status.gameOver, 0, Item(2, 16, 1), ..., Vec2d(0, -1), Board(10, 20))
            ]
```

If I want to test only `status` return, most of the info in `Actual` and `Expected` are no useful.
IMHO, the missing of `mapState` is a pain point in `blocTest`.

---

With an map, I can do something like: 

```
    blocTest(
      'teste',
      ...
      stateMap: (state) => state.status,
      expect: [Status.running],
    );
```

felangel:
I’m not sure I understand why you would want to test for a certain property instead of the entire state. That could lead to false positives especially if you use subclasses to represent different states.

lucasnlm:
And it will compare only the fields of the state that I want to test. If any other field is different and it's OK be different, the test won't fail.

felangel:
But in that case you’re no longer truly testing the bloc. It also doesn’t really work in cases where you use subclasses to represent your bloc states

lucasnlm:
Yeap, to test with subclasses It would be necessary to test the instance of `(state)`. But, since it's an optional field it can still be used as is today.

---

About your first point. I think it depends on what you want to test. Not necessarily an event will change the whole state. If something else broke something in the bloc, an other test must catch it.

In my example: I want to test the `status` from a game after some events. So, I would like to check only the `status` on `expect`. If I have `items`, `points` and `etc` in the state, they can be tested on specific tests for them.

I think it may also be done with a custom `Matcher` (I'm not sure). But It would be more boilerplate.

felangel:
@lucasnlm since this can already be accomplished with custom matchers and isn't that much boilerplate I would prefer not to complicate the blocTest api (especially when it only makes sense for a specific type of state).

You can write a custom matcher for your case like:
```dart
Matcher hasStatus(Status status) => TypeMatcher<MyState>()
    .having((i) => i.status, 'has correct status', status);
```

Then in all your tests you can just have

```dart
blocTest(
  'test running',
  ...
  expect: [hasStatus(Status.running)],
);

blocTest(
  'test walking',
  ...
  expect: [hasStatus(Status.walking)],
);
```

lucasnlm:
That's ok. I still think that a `map` is simpler and emits a result easier to read than a `Matcher`.
This PR can be closed. 👍 
Thank you for your time and good job with bloc. 😄

felangel:
Here's a complete example

```dart

enum Status { running, walking }

class MyState {
  const MyState(this.status);

  final Status status;

  @override
  String toString() => 'MyState($status)';
}

class MyBloc extends Bloc<Object, MyState> {
  @override
  MyState get initialState => MyState(Status.walking);

  @override
  Stream<MyState> mapEventToState(Object event) async* {
    yield MyState(Status.running);
  }
}

Matcher hasStatus(Status status) =>
    TypeMatcher<MyState>().having((i) => i.status, '$status', status);

void main() {
  blocTest(
    'test',
    build: () async => MyBloc(),
    act: (bloc) => bloc.add(Object),
    expect: [
      hasStatus(Status.walking),
    ],
  );
}
```

That test fails with the following exception:

```sh
Expected: [<<Instance of 'MyState'> with `Status.walking`: Status:<Status.walking>>]
  Actual: [MyState:MyState(Status.running)]
   Which: does not match <Instance of 'MyState'> with `Status.walking`: Status:<Status.walking> at location [0]
```

While the error is slightly more verbose than the one produced in this PR, I feel it's still easy to see why the test failed.

I would love to see a concrete use case for having a mapper or using a custom matcher (unless you're not overriding `==` in your states) because in my opinion it's best to always write `blocTest` with the entire state in mind.

PR #1069, opened 2020-04-20
surface unhandled bloc errors in debug mode

felangel:
## Status
**READY**

## Breaking Changes
YES

## Description
- In debug mode, a `BlocUnhandledErrorException` will be thrown to make errors apparent while in development.
- In release mode, the bloc's behavior will remain unchanged.

As a result, with these changes it becomes important to call `super.onError` last when overriding `onError` in order to ensure the custom `onError` code is executed.

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

- Breaking changes which will be published in bloc v4.0.0-dev.4

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -4,6 +4,28 @@ import 'package:meta/meta.dart';
 
 import '../bloc.dart';
 
+/// {@template bloc_unhandled_error_exception}
+/// Exception thrown in debug mode when an unhandled error occurs within a bloc.
+/// {@endtemplate}
+class BlocUnhandledErrorException implements Exception {
+  /// The [bloc] in which the unhandled error occurred.
+  final Bloc bloc;
+
+  /// The unhandled [error] object.
+  final Object error;
+
+  /// An optional [stackTrace] which accompanied the error.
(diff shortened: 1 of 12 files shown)


packages/bloc/lib/src/bloc.dart
```diff
@@ -4,6 +4,28 @@ import 'package:meta/meta.dart';
...
+  final Bloc bloc;
+
+  /// The unhandled [error] object.
+  final Object error;
+
+  /// An optional [stackTrace] which accompanied the error.
+  final StackTrace stackTrace;
+
+  /// {@macro bloc_unhandled_error_exception}
+  BlocUnhandledErrorException(this.bloc, this.error, this.stackTrace);
```
orsenkucher, line 21: make **[** this.stackTrace **]** optional?

packages/bloc/lib/src/bloc.dart
```diff
@@ -4,6 +4,28 @@ import 'package:meta/meta.dart';
...
+
+  @override
+  String toString() {
+    return 'Unhandled error $error occurred in bloc $bloc.\n$stackTrace';
```
orsenkucher, line 25: ${stackTrace??"No stack trace available."}

PR #1228, opened 2020-05-30
flutter_bloc: BlocProvider's `create` precursors

orsenkucher:
## Status
**READY**

## Breaking Changes
NO

## Description
We'll probably need this change to make `package:hydrated_bloc` storage scopes convenient to use 
So I'll file this change in a PR🥰

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_provider.dart
+++ b/packages/flutter_bloc/lib/src/bloc_provider.dart
@@ -100,7 +100,7 @@ class BlocProvider<T extends Bloc<dynamic, dynamic>>
   /// ```dart
   /// BlocProvider.of<BlocA>(context)
   /// ```
-  static T of<T extends Bloc<dynamic, dynamic>>(BuildContext context) {
+  static T of<T extends Bloc>(BuildContext context) {
     try {
       return Provider.of<T>(context, listen: false);
     } on ProviderNotFoundException catch (_) {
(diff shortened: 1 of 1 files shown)


felangel:
@orsenkucher can you provide a bit of context regarding why this change is necessary and what you’re envisioning the HydratedBloc scoped api would look like? Thanks 👍

orsenkucher:
🙌 @felangel 
HydratedBloc needs to get information about a scope it was created in somehow🤷 
Passing it via constructor is tedious, so I came up with a solution to extract required info from context inside `BlocProvider.build` method, and then a moment before bloc instantiation add this token to queue
```dart
BlocProvider.registerPrecursor((context, create) {
  final token = context.scope().token;
  return (context) {
    Hydrated.register(token);
    return create(context);
  };
});
```
and basically on next execution task pick it up like so
```dart
abstract class HydratedBloc<Event, State> extends Bloc<Event, State> {
  HydratedBloc() : _scopeToken = Hydrated.acquire() {
    ...
  }
}
```

Also, I'm logging ideas in this [issue](https://github.com/felangel/hydrated_bloc/issues/54)

orsenkucher:
> api would look like?

I envision it to be as stealth and implicit as possible
Without crazy stuff to get it working on user side😄

orsenkucher:
But making it familiar and pragmatic as well

felangel:
Going to close this for now since there have been a lot of changes recently and I want to think about the best way to approach this. I'll use this PR for inspiration and will let you know once I have some sort of proposal, thanks! 🙏

PR #1302, opened 2020-06-13
feat!: v5.0.0

### LINKED ISSUE #1236: [Feature request] Events added after event controller was closed should not result in an error but instead should be ignored
Adding events from a delayed future could very easily result in an error if the bloc is closed before this future has added the event. This type of scenario forces the developer to have to manually cancel the future to ensure it won't result in an error.

An event added after the event controller was closed should be ignored and not result in an error, since trying to add an event should not be considered an error just because the bloc was closed.

```
@override
  void add(Event event) {
    try {
      if (_eventController.isClosed) {
        // event was added after closing event controller, so it will be ignored
        return;
      }
      onEvent(event);        
       _eventController.add(event);
    } on dynamic catch (error, stackTrace) {
      onError(error, stackTrace);
    }
  }

```

This would ensure only actual errors would end up in the `onError`.

Manually cancelling the future, but that's tedious, repetitive and time consuming. 

**Additional context**
None.

### LINKED ISSUE #1304: [BREAKING][Proposal] remove initialState override
As a developer, having to override `initialState` when creating a bloc presents two main issues:

1. The `initialState` of the bloc can be dynamic and can also be referenced at a later point in time (even outside of the bloc itself). In some ways, this can be viewed as leaking internal bloc information to the UI layer.
2. It's verbose.

Rather than defining the initial state like:

```dart
class CounterBloc extends Bloc<CounterEvent, int> {
  @override
  int get initialState => 0;

  ...
}
```

I am proposing:

```dart
class CounterBloc extends Bloc<CounterEvent, int> {
  CounterBloc() : super(0);

  ...
}
```

This will avoid exposing `initialState` publicly and also reduces the amount of code needed especially considering in most cases a bloc will already require a constructor for injecting a repository like:

```dart
class MyBloc extends Bloc<MyEvent, MyState> {
  MyBloc(this._myRepository);

  final MyRepository _myRepository;

  @override
  MyState get initialState => MyState.initial();
}
```

The above case could be simplified to:

```dart
class MyBloc extends Bloc<MyEvent, MyState> {
  MyBloc(this._myRepository) : super(MyState.initial());

  final MyRepository _myRepository;
}
```

Additionally, it would no longer be possible to dynamically modify the value of a bloc's `initialState` after it has been initialized.

In the current state, the following code is totally valid but could potentially have many unintended side-effects

```dart
class CounterBloc extends Bloc<CounterEvent, int> {
  @override
  int get initialState => _initialState;

  int _initialState = 0;

  @override
  Stream<int> mapEventToState(CounterEvent event) async* {
    switch (event) {
      case CounterEvent.increment:
        _initialState = 100; // this should ideally not be possible
        yield state + 1;
        break;
      default:
        break;
    }
  }
}
```

**Additional context**
This would also enable reduction in boilerplate for `HydratedBloc`

Before:

```dart
class CounterBloc extends HydratedBloc<CounterEvent, int> {
  @override
  int get initialState => super.initialState ?? 0;

  ...
}
```

We can simplify the API and eliminate the need to explicitly call `super.initialState`

```dart
class CounterBloc extends HydratedBloc<CounterEvent, int> {
  CounterBloc() : super(0);

  ...
}
```

Would love to hear everyone's feedback on this especially if you are specifically relying on `initialState` for something outside of just seeding the bloc's initial state 👍 

Looking forward to hearing everyone's thoughts 😄 

**UPDATE**
After going through a lot of feedback it seems a good alternative would be to use a named parameter in `super`.

```dart
class CounterBloc extends Bloc<CounterEvent, int> {
  CounterBloc() : super(initialState: 0);

  ...
}
```

### LINKED ISSUE #1312: It is not possible for state to be null
**Describe the bug**
It seems that at the moment `bloc` does not allow the state (initial or current) to be null. If either of them are null, this error occurs: `Unhandled error 'package:bloc/src/transition.dart': Failed assertion: line 25 pos 16: 'currentState != null': is not true.`.

**To Reproduce**
1. Create a Bloc component with an initial state of `null`.
2. Yield any object in `mapEventToState`.
3. See error

**Expected behavior**
I expect to be able to use null as a valid current state.

felangel:
## Status
**READY**

- [X] bloc v5.0.0-dev.11
- [X] flutter_bloc v5.0.0-dev.5
- [X] angular_bloc v6.0.0-dev.4
- [X] bloc_test v6.0.0-dev.4

## Breaking Changes
YES

## Description
- Refactor bloc to extend [cubit](https://github.com/felangel/cubit) rather than `Stream`. (non-breaking)
- Support `null` states (closes #1312)
- Ignore newly added events after bloc is closed (closes #1236)
- **BREAKING**: Removing `initialState` override in favor of providing the initial state via `super`. (closes #1304)

```dart
class CounterBloc extends Bloc<CounterEvent, int> {
  @override
  int get initialState => 0;
}
```

would become

```dart
class CounterBloc extends Bloc<CounterEvent, int> {
  CounterBloc() : super(0);
}
```

The other consequence of this, `initialState` will no longer exist as a public getter on blocs so I would love to hear what everyone thinks about this change. If the community prefers to keep the `initialState` signature as is for now, I'm definitely open to leaving it as is but feel it's a bit clunky and could be streamlined.

- **BREAKING**: remove `BlocSupervisor` and rename `BlocDelegate` to `BlocObserver`

```dart
class SimpleBlocDelegate extends BlocDelegate {
 @override
  void onEvent(Bloc bloc, Object event) {
    print('bloc: ${bloc.runtimeType}, event: $event');
    super.onEvent(bloc, event);
  }

  @override
  void onTransition(Bloc bloc, Transition transition) {
    print('bloc: ${bloc.runtimeType}, transition: $transition');
    super.onTransition(bloc, transition);
  }

  @override
  void onError(Bloc bloc, Object error, StackTrace stackTrace) {
    print('bloc: ${bloc.runtimeType}, error: $error');
    super.onError(bloc, error, stackTrace);
  }
}
```

would become:

```dart
class SimpleBlocObserver extends BlocObserver {
  @override
  void onEvent(Bloc bloc, Object event) {
    print('bloc: ${bloc.runtimeType}, event: $event');
    super.onEvent(bloc, event);
  }

  @override
  void onTransition(Bloc bloc, Transition transition) {
    print('bloc: ${bloc.runtimeType}, transition: $transition');
    super.onTransition(bloc, transition);
  }

  @override
  void onError(Bloc bloc, Object error, StackTrace stackTrace) {
    print('bloc: ${bloc.runtimeType}, error: $error');
    super.onError(bloc, error, stackTrace);
  }
}
```

And initialization would go from:

```dart
BlocSupervisor.delegate = SimpleBlocDelegate();
```

to

```dart
Bloc.observer = SimpleBlocObserver();
```

- Bloc to implement `EventSink` rather than `Sink` which exposes `addError` for a more intuitive, friendly API to handle exceptions within a bloc.

Currently you have to `rethrow`:

```dart
Stream<MyState> mapEventToState(MyEvent event) async* {
  try {
    await someOperation();
  } on SomeSpecificException catch (_) {
    yield SomeSpecificFailureState();
    rethrow; // currently necessary to propagate error to `onError` in `BlocObserver`.
  }
}
```

Or `throw` when using `Either`:

```dart
yield* either.fold(
  (failure) async * {
    yield FailureState(failure);
    throw failure;
  }, 
  (data) => ...
)
```

It is not immediately clear/obvious that (re)throwing bubbles the exception up to `onError` and can be improved:

```dart
Stream<MyState> mapEventToState(MyEvent event) async* {
  try {
    await someOperation();
  } on SomeSpecificException catch (error, stackTrace) {
    yield SomeSpecificFailureState();
    addError(error, stackTrace);
  }
}
```

```dart
yield* either.fold(
  (failure) async * {
    yield FailureState(failure);
    addError(failure);
  }, 
  (data) => ...
)
```

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

- Breaking Changes introduced in bloc v5.0.0


zepfietje:
@felangel could you elaborate on the motive for this change (making bloc extend cubit)? Have you already got a clearer idea of when to use cubit vs. bloc?

The `initialState` change does seem clean to me. :ok_hand:

RollyPeres:
> IMO, the goal of a constructor should be just to build an instance, but in this case is building the instance plus setting the first/initial value, which I see a bit more confusing than specifying directly `initialState`.

@jorgecoca  that could be said about any class which has fields since you have to initialize them; so basically you're building the instance and setting initial values through the constructor.

```
class Model {
  Model._(): field = '';
  const Model(this.field);

  final String field;
}
```

Even with the current bloc implementation the initialization happens within the constructor, just routed through a different property.

```
 Bloc() {
    _state = initialState;
    _bindEventsToStates();
  }
```

Dart team also insists on not using properties just for the sake of being fancy when initializing a field, but rather initialize the field directly.

I personally like this change, especially in regards to `HydratedBloc`.

felangel:
@zepfietje I'm working on a formal proposal/roadmap for cubit and will share it as soon as I can.
@jorgecoca as @RollyPeres mentioned, I personally think the proposed change actually makes more sense because the initialState should not be something that can change after the bloc has been instantiated and as was pointed out, the underlying implementation isn't affected by this change (we're still seeding the `_state` of the bloc in both cases). Would be good to move this discussion to #1304 for visibility 👍

packages/bloc/lib/src/bloc.dart
```diff
@@ -43,7 +43,9 @@ abstract class Bloc<Event, State> extends Cubit<State> implements Sink<Event> {
   StreamSubscription<Transition<Event, State>> _transitionSubscription;
 
   /// {@macro bloc}
-  Bloc(State state) : super(state) {
+  Bloc({@required State initialState})
+      : assert(initialState != null),
```
RollyPeres, line 47: What if the user wants to explicitly start with a null state ? e.g.: bloc's state is of type `String` and the user wants `: super(initialState: null)`

felangel: Is that a use case we want to support? I never intended for blocs/cubits to have null states.

RollyPeres: I wouldn't encourage it, but there's surely a chance someone might need it. I feel like we could allow for it especially since null safety is quickly approaching and people might want to have a state like `State? `. It would feel like we're removing with feature. Also `initialState` being marked as `@required` means the user will have to explicitly set it to `null` if that's what is needed, so no `null` `initialState` surprise there.

felangel: Yeah that's definitely a good point. Thanks for brining that up 👍

RollyPeres: > There's also currently nothing preventing a developer from yielding a null state so it wouldn't be consistent anyway.

Great point 👍

felangel:
@jorgecoca @RollyPeres @chimon2000 can you take another look when you have a min and lmk if you have any additional feedback that I haven't incorporated? Based on the discussions in #1304 it seems safe to proceed. Thanks 👍

packages/bloc_test/README.md
```diff
@@ -1,4 +1,4 @@
-<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/bloc_test_logo_full.png" height="100" alt="Bloc Test Package" />
+<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/bloc_test_logo_full.png" height="100px" alt="Bloc Test Package" />
```
zepfietje, line 1: @felangel, unit should be removed.
```suggestion
<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/bloc_test_logo_full.png" height="100" alt="Bloc Test Package" />
```

packages/bloc_test/README.md
```diff
@@ -194,3 +194,7 @@ test('emits [1] when CounterEvent.increment is added', () async {
 ## Maintainers
 
 - [Felix Angelov](https://github.com/felangel)
+
+## Supporters
+
+[<img src="https://raw.githubusercontent.com/felangel/bloc/master/docs/assets/vgv_logo.png" width="120px" />](https://verygood.ventures)
```
zepfietje, line 200:
```suggestion
[<img src="https://raw.githubusercontent.com/felangel/bloc/master/docs/assets/vgv_logo.png" width="120" />](https://verygood.ventures)
```

packages/angular_bloc/README.md
```diff
@@ -1,4 +1,4 @@
-<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/angular_bloc_logo_full.png" height="150" alt="Angular Bloc Package" />
+<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/angular_bloc_logo_full.png" height="100px" alt="Angular Bloc Package" />
```
zepfietje, line 1:
```suggestion
<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/angular_bloc_logo_full.png" height="100" alt="Angular Bloc Package" />
```

packages/bloc/README.md
```diff
@@ -21,7 +21,7 @@ This package is built to work with [flutter_bloc](https://pub.dev/packages/flutt
 
 ## Overview
 
-<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/bloc_architecture.png" width="250px" alt="Bloc Architecture" />
+<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/bloc_architecture.png" width="300px" alt="Bloc Architecture" />
```
zepfietje, line 24:
```suggestion
<img src="https://raw.githubusercontent.com/felangel/bloc/feat/bloc-extends-cubit/docs/assets/bloc_architecture.png" width="300" alt="Bloc Architecture" />
```

packages/bloc_test/README.md
felangel: @zepfietje why do you prefer to avoid the units?

packages/bloc/README.md
```diff
@@ -201,4 +201,4 @@ At this point, all `Bloc` `Exceptions` will also be reported to the `SimpleBlocD
 
 ## Supporters
 
-[<img src="https://raw.githubusercontent.com/felangel/bloc/master/docs/assets/vgv_logo.png" style="max-width:120px" />](https://verygood.ventures)
+[<img src="https://raw.githubusercontent.com/felangel/bloc/master/docs/assets/vgv_logo.png" width="120px" />](https://verygood.ventures)
```
zepfietje, line 204:
```suggestion
[<img src="https://raw.githubusercontent.com/felangel/bloc/master/docs/assets/vgv_logo.png" width="120" />](https://verygood.ventures)
```

packages/bloc_test/README.md
zepfietje: @felangel https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width

felangel: oops! I was thinking of the width/height css styles 😛

zenkog:
I think this is a good change. Got some minor code change to do though. But it's okay. Great work @felangel !!! Thank you !! Lots of Thumbs Up!!

PR #1322, opened 2020-06-20
update rxdart to 0.24.0

kateile:
## Status
**READY**

## Breaking Changes
NO

## Description
I face dependencies conflicts.  Other dependencies I use use latest version of rxdart.


felangel:
Hi @kateile 👋 
Thanks for opening an PR! I'll merge this but I don't think it should be a problem either way because bloc does not depend on rxdart (the dev_dependency is just for tests) so you shouldn't have an issues using the latest version of rxdart with bloc. Can you please share a link to a sample app which illustrates the problem? Are you using the latest version of bloc? Thanks 👍

PR #1357, opened 2020-06-30
feat(bloc): do not throw in onError with custom observer

### LINKED ISSUE #1254: Mark certain exceptions as handled in onError()
Just looking at https://github.com/felangel/bloc/issues/1122

What about the following design pattern:

I am handling the API errors throughout the bloc's in the BlocDelegate.onError() method by showing a message to the user to avoid endless try... catch boilerplate in the blocs.

A lot of api blocs error handling is as simple as showing the error message and stop at the point of the exception, and if something extra is needed it can be handled in the bloc's onError.

It would be nice if I could mark those exceptions as handled to avoid unnecessary `unhandled exceptions` messages in the debug mode.

Would a simple `ignore: true` extra argument to the super onError work for this case?

example, what i'd like to be able to do in the delegate:

```
  @override
  void onError(Bloc bloc, Object error, StackTrace stacktrace) {
    messageBloc?.add(MessageError(exception: error));
    super.onError(bloc, error, stacktrace, ignore: true);
  }

```

Well in this case I actually don't really want to call `super.onError` but as it is no longer allowed, what would be a better solution if not the above?

felangel:
## Status
**READY**

## Breaking Changes
NO

## Description
- When custom `BlocObserver` is provided, avoid throwing `BlocUnhandledErrorException` (closes #1254)

## Todos
- [X] Tests
- [X] Documentation
- [X] Examples

- To be included in v5.0.0 release of bloc

### DIFF
--- a/packages/bloc/CHANGELOG.md
+++ b/packages/bloc/CHANGELOG.md
@@ -1,10 +1,14 @@
+# 5.0.0-dev.11
+
+- feat: do not throw `BlocUnhandledException` when custom `BlocObserver` is used ([#1254](https://github.com/felangel/bloc/issues/1254)).
+
 # 5.0.0-dev.10
 
-- docs: additional minor improvement to bloc logo alignment
+- docs: additional minor improvement to bloc logo alignment.
 
 # 5.0.0-dev.9
 
-- docs: minor improvement to bloc logo alignment
+- docs: minor improvement to bloc logo alignment.
 
 # 5.0.0-dev.8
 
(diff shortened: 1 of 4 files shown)


felangel:
After several discussions I'm closing this because `BlocObserver` should be a passive observer which could log/track state changes or errors but should not contain business logic or manipulate the UI. If you want to have global error handling (in which you show an overlay or dialog) I would recommend handling the exceptions at the repository layer or having an http interceptor (for low level network errors) which can all be consumed by a global bloc rather than having each bloc intentionally throw uncaught exceptions and leave it up to the BlocObserver to "handle" them.

PR #1359, opened 2020-06-30
extend bloc_test

### LINKED ISSUE #1358: Test async Expect values like image fixtures
Currently the expect field takes an iterable. I have an imageFixture that loads a local file and I want to use this fixture in my expect statement. 

I extended the bloc_test to take an optional expectAsync field which takes an async function returning Future<Iterable>

Using normal tests. 

**Additional context**

lunaticcoding:
## Status
READY

## Breaking Changes
NO

## Description
Fixes #1358

## Related PRs
None

## Steps to Test or Reproduce

```
```

This PR will affect:

*

### DIFF
--- a/packages/bloc_test/test/bloc_test_test.dart
+++ b/packages/bloc_test/test/bloc_test_test.dart
@@ -177,7 +177,7 @@ void main() {
           await Future.delayed(Duration(milliseconds: 10));
           bloc.add(MultiCounterEvent.increment);
         },
-        expect: () async => [1, 2, Future.value(3), 4],
+        expectAsync: () async => [1, 2, Future.value(3), 4],
       );
     });
 
--- a/packages/bloc_test/test/bloc_test_test.dart
+++ b/packages/bloc_test/test/bloc_test_test.dart
@@ -53,16 +53,20 @@ void main() {
         expect: [1, 2],
       );
 
+      int asyncNumber;
       blocTest(
         'emits [1, 2] when CounterEvent.increment is called multiple times'
-        'with async act and expectAsync',
-        build: () async => CounterBloc(),
+        'with async act and expectAsFunction',
+        build: () async {
+          asyncNumber = await Future.value(2);
+          return CounterBloc();
+        },
         act: (bloc) async {
           bloc.add(CounterEvent.increment);
           await Future.delayed(Duration(milliseconds: 10));
           bloc.add(CounterEvent.increment);
         },
-        expectAsync: () async => [1, await Future.value(2)],
+        expectAsFunction: () => [1, asyncNumber],
       );
     });
 
@@ -97,18 +101,6 @@ void main() {
           bloc.add(AsyncCounterEvent.increment);
         },
         expect: [1, 2],
-      );
-
-      blocTest(
-        'emits [1, 2] when AsyncCounterEvent.increment is called multiple'
-        'times with async act and expectAsync',
-        build: () async => AsyncCounterBloc(),
-        act: (bloc) async {
-          bloc.add(AsyncCounterEvent.increment);
-          await Future.delayed(Duration(milliseconds: 10));
-          bloc.add(AsyncCounterEvent.increment);
-        },
-        expectAsync: () async => [Future.value(1), 2],
       );
     });
 
@@ -166,18 +158,6 @@ void main() {
           bloc.add(MultiCounterEvent.increment);
         },
         expect: [1, 2, 3, 4],
-      );
-
-      blocTest(
-        'emits [1, 2, 3, 4] when MultiCounterEvent.increment is called'
-        'multiple times with async act and expectAsync',
-        build: () async => MultiCounterBloc(),
-        act: (bloc) async {
-          bloc.add(MultiCounterEvent.increment);
-          await Future.delayed(Duration(milliseconds: 10));
-          bloc.add(MultiCounterEvent.increment);
-        },
-        expectAsync: () async => [1, 2, Future.value(3), 4],
       );
     });
 
--- a/packages/bloc_test/test/bloc_test_test.dart
+++ b/packages/bloc_test/test/bloc_test_test.dart
@@ -52,6 +52,18 @@ void main() {
         },
         expect: [1, 2],
       );
+
+      blocTest(
+        'emits [1, 2] when CounterEvent.increment is called multiple times'
+        'with async act and expectAsync',
+        build: () async => CounterBloc(),
+        act: (bloc) async {
+          bloc.add(CounterEvent.increment);
+          await Future.delayed(Duration(milliseconds: 10));
+          bloc.add(CounterEvent.increment);
+        },
+        expectAsync: () async => [1, await Future.value(2)],
+      );
     });
 
     group('AsyncCounterBloc', () {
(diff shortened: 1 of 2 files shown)


felangel:
Hi @lunaticcoding 👋 
Thanks for opening an PR. Can you please provide an example which illustrates the current limitation of blocTest? I think expect should also handle async expectations internally and we could have a single expect to handle both cases.

lunaticcoding:
@felangel the limitation I encountered was the following:

```
blocTest(
"ChangeImage changes the image",
    build: createBloc,
    act: (bloc) => bloc.add(
      DatabaseInterfaceEvent.imageChanged(
        await imageFixture("img.png"),
      ),
    ),
    expectAsync: () async => [
      DatabaseInterfaceState.initial().copyWith(
        type: ValidatedImage(
          await imageFixture("img.png"),
        ),
      ),
    ],
  );
```

with the code crashing because ValueObjects have an assert(!= null). So even when I try the following, the assert gets triggered:
```
UintList8 newImage;
blocTest(
"ChangeImage changes the image",
    build: () async {
        newImage = await imageFixture("img.png");
        return createBloc();
    },
    act: (bloc) => bloc.add(
      DatabaseInterfaceEvent.imageChanged(
        newImage,
      ),
    ),
    expectAsync: () async => [
      DatabaseInterfaceState.initial().copyWith(
        type: ValidatedImage(
          newImage,
        ),
      ),
    ],
  );
```

A possible workaround would be to initialise the newImage in a setUp/setUpAll method, but I would really like to not have that many global variables floating around. (The state is quite big and there is quite a few fixtures to test for).

felangel:
@lunaticcoding can you please share a link to a github repo which reproduces the issue without the `expectAsync`? It would really help me get some more context 🙏

lunaticcoding:
@felangel I unfortunately can't share this code (I'll ask my boss if it's okay if I share just the ValidationImage and the complete bloc if that helps), but I can set up something similar in a public repo over the weekend. :)

packages/bloc_test/test/bloc_test_test.dart
```diff
@@ -86,6 +98,18 @@ void main() {
...
+      blocTest(
+        'emits [1, 2] when AsyncCounterEvent.increment is called multiple'
+        'times with async act and expectAsync',
+        build: () async => AsyncCounterBloc(),
+        act: (bloc) async {
+          bloc.add(AsyncCounterEvent.increment);
+          await Future.delayed(Duration(milliseconds: 10));
+          bloc.add(AsyncCounterEvent.increment);
+        },
+        expectAsync: () async => [Future.value(1), 2],
```
felangel, line 111: I don't really understand this. The bloc should not be emitting Futures can you please explain why this is needed?

packages/bloc_test/test/bloc_test_test.dart
```diff
@@ -143,6 +167,18 @@ void main() {
...
+      blocTest(
+        'emits [1, 2, 3, 4] when MultiCounterEvent.increment is called'
+        'multiple times with async act and expectAsync',
+        build: () async => MultiCounterBloc(),
+        act: (bloc) async {
+          bloc.add(MultiCounterEvent.increment);
+          await Future.delayed(Duration(milliseconds: 10));
+          bloc.add(MultiCounterEvent.increment);
+        },
+        expectAsync: () async => [1, 2, Future.value(3), 4],
```
felangel, line 180: same here

felangel:
> @felangel I unfortunately can't share this code (I'll ask my boss if it's okay if I share just the ValidationImage and the complete bloc if that helps), but I can set up something similar in a public repo over the weekend. :)

Thanks yeah a super simple example would be great because in the current state I'm not sure I understand the need for this change.

packages/bloc_test/test/bloc_test_test.dart
lunaticcoding: to test if the future returns before comparing the expected array with the actual states.

felangel: but the way `blocTest` works is it will wait until all the states are emitted before calling `expect`

lunaticcoding:
@felangel I created a very bare-bone project to show what I mean :) 
It's here :) https://github.com/lunaticcoding/bloc_test_demo

PS: If you are not familiar with value objects (https://www.youtube.com/watch?v=fdUwW0GgcS8)

felangel:
> @felangel I created a very bare-bone project to show what I mean :)
> It's here :) https://github.com/lunaticcoding/bloc_test_demo
> 
> PS: If you are not familiar with value objects (https://www.youtube.com/watch?v=fdUwW0GgcS8)

Thanks for the example! So the root cause is that expect returns an `Iterable` not a lambda so in your case `asyncNumber` is `null` in the `expect` because it is evaluated immediately. The way to fix this would be to change `expect` to be a `Iterable Function()` which would be a really nasty breaking change.

I feel like it would be best to avoid loading async data in a test. Is it absolutely necessary or can you instead work with constants or local values?

```dart
void main() {
  const value = 12; // if it must be image data just reference a constant `UInt8List` in a `test_fixtures.dart` 

  blocTest(
    '...',
    build: () async => MyBlocBloc(),
    act: (bloc) async => bloc.add(MyConcreteBlocEvent(value)),
    expect: [MyBlocSecondState(ConcreteValueObject(value))],
  );
```

Let me know what you think 😄

lunaticcoding:
@felangel well the problem is that I need to test on images. So I use 

```
Future<Uint8List> fileFixture(String name) async =>
    (await File('test/fixtures/$name').readAsBytes()).buffer.asUint8List();
```

to load an image async and then I want to test whether my bloc creates the correct value object. :S I think my workaround would be to just default to regular tests for those cases.

felangel:
> @felangel well the problem is that I need to test on images. I would prefer not to have the entire fixture in a dart file because it will just be messy ;D So I use
> 
> ```
> Future<Uint8List> fileFixture(String name) async =>
>     (await File('test/fixtures/$name').readAsBytes()).buffer.asUint8List();
> ```
> 
> to load an image async and then I want to test whether my bloc creates the correct value object. I have an image in the same folder as my fixture reader (just a dart file holding this function and a similar one for json files).
> 
> :S I think my workaround would be to just default to regular tests for those cases.
> 
> Another suggestion would be to add a expectFromFunction/expectAsFunction. This would then give the option to load a Iterable Function() (async stuff can be bypassed by initialising it in the build method), without any breaking changes.

Instead of storing your fixtures like that why not just store the `Uint8List` directly?

```dart
// get the Uint8List representation of your image and storage that instead of the image.
// this will both save you time when running tests since you don't need to read a file each time
// and also make your tests safe since you don't rely on external functionality like readAsBytes etc...
final fixture1 = Uint8List.fromList([0, 1, 2, 3,]);
```

felangel:
@lunaticcoding thoughts? Going to close this PR for now since I feel the proposed API would cause confusion and I would like to avoid changing the `expect` method signature unless it's absolutely necessary.

lunaticcoding:
@felangel I don't really like having to copy my fixtures into a file (I would have to do something weird like reading it in, copying it into a new dart file) This would not remove the external dependency but merely hide it. 

I can simply fall back to writing a more verbose plain old test test instead of a bloc_test test ;D 
But thanks anyways.

felangel:
Flutter does this as well https://github.com/flutter/flutter/blob/462b0ea76eca028a7776f60a20a05ebce30dfdbd/packages/flutter/test/widgets/image_data.dart#L7 and it’s a one time thing that is used for testing so I highly recommend that approach over actually reading the file.

Closing for now but happy to continue the conversation 👍

lunaticcoding:
@felangel Oh wow. I'll adapt my approach then :D thanks for taking the time to look this up for me :) really appreciate it! Cheers.

PR #1361, opened 2020-06-30
Add getter isClosed for bloc

ZaarU:
## Status
**READY**

## Breaking Changes
NO

## Description
Add getter isClosed for bloc by using 
  bool get isClosed => _eventController.isClosed;


felangel:
Hi @ZaarU 👋 
Thanks for opening an issue!

Can you please provide some context for why this change is necessary? Thanks 👍

packages/bloc/lib/src/bloc.dart
```diff
@@ -53,6 +53,9 @@ abstract class Bloc<Event, State> extends Stream<State> implements Sink<Event> {
   @override
   bool get isBroadcast => _stateController.stream.isBroadcast;
 
+  /// Returns that the [bloc] was closed.
```
felangel, line 56:
```suggestion
  /// Returns whether the [Bloc] is closed.
```

ZaarU:
` @override
  Stream<SomeState> mapEventToState(
    SomeEvent event,
  ) async* {
    if (event is NeedSomeDataEvent) {
      _getDataAsync();
    }
  }

  _getDataAsync() async {
    var someData = await _repository.getBitLongData();  // data goes a long time

  /// if widget is closed, here we must catch
  /// exception or check if it was closed.
    add(DataReceivedEvent(someData));   
  }`

felangel:
@ZaarU in what case would you be uncertain whether the bloc is closed? `add` does nothing by default if the bloc is already closed (in v5.0.0).

ZaarU:
> @ZaarU in what case would you be uncertain whether the bloc is closed? `add` does nothing by default if the bloc is already closed (see #1236). it will be included in v5.0.0 and is available in 5.0.0-dev currently

Thanks.
Sorry, i missed this PR. Closed

PR #1526, opened 2020-07-25
feat: hydrated storage scope

orsenkucher:
## Status

**READY TO BE REVIEWED**

## Breaking Changes

NO

## Description

- Added `CreateWrapper` to flutter_bloc.
- Refactored `HydratedMixin` to use `this.instanceStorage`.
- Added `HydratedScope` - READY FOR REVIEW.
- [Modified](https://github.com/orsenkucher/bloc/pull/2/files) counter example with secure storage toggle button. 

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [x] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #1559, opened 2020-07-31
feat(flutter_bloc): add cubit widgets

jeroen-meijer:
## Status

**WORK IN PROGRESS**

## Breaking Changes

YES

## Description

This PR adds `Cubit` equivalents for all `Bloc` widgets and refactors all `Bloc` widgets to use a `bloc` parameter instead of a `cubit` parameter (similar to the v5 API, see `migration.md`). (#1560)

### Reasoning

When `bloc` got upgraded to use cubits internally in v6.0.0, and `flutter_bloc` followed suit, all `Bloc` widgets (such as `BlocBuilder`, `BlocConsumer` and `BlocListener`) had their APIs changed and the `bloc` parameter was rightfully rename `cubit`. This was a breaking change in the API.

However, through trying out and experimenting v6, I found it confusing to provide `BlocBuilder`s and similar widgets with a `cubit` instead of a `bloc` parameter, especially considering I had never heard of or used a `Cubit` before.

> _What are these newfangled cew-bids those darned kids keep talkin' 'bout?_

Similarly, it's confusing to me to have to use a `Bloc` widget for `cubit`s. Seeing a `BlocBuilder` makes me instantly/instinctly think there is a `Bloc` somewhere I need to keep track of, even when there isn't and in reality I'm using a `cubit`.

Ultimately, this is all a preference thing. However, I think code would read much more nicely if `BlocBuilder`s contained `bloc`s, and `CubitBuilder`s contained `cubit`s. Having to import a library called `flutter_bloc` instead of `flutter_cubit` is totally fine with me, though, since they're so closely related and the chance of somebody using both in the same library/application is pretty high.

### Note

This is a controversial change, and it may be too early for people to adopt and change back their widgets to how it was in the previous major version.

However, in my opinion, considering the API is now very similar to the way it was in v5, these changes can be considered non-breaking for people who are using anything but v6 (currently the latest version).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


orsenkucher:
Hi @jeroen-meijer 
Can you please share the knowledge how to get Codecov Report message😄?

jeroen-meijer:
> Hi @jeroen-meijer
> Can you please share the knowledge how to get Codecov Report message😄?

Check https://codecov.io/

orsenkucher:
> Check https://codecov.io/

Yeah, I mean I have #1526 in bloc repo, but it's without codecov report message, which is very convenient to have

orsenkucher:
@felangel I'm almost sure https://github.com/felangel/bloc/pull/1559#issuecomment-667503367 was posted only after all other checks have passed. Maybe we should consider removing code coverage from github-actions CI, as it's available as separate check?

felangel:
@jeroen-meijer based on the feedback on #1560 thoughts on closing this for now? We can always reopen it or an adjusted version once we reach a consensus on the direction/approach 👍

felangel:
@jeroen-meijer closing this for now based on the response on #1560. I really appreciate the time you took to do this and we can definitely reopen this PR once we have a better sense of how the API will evolve in the future 👍

Issue #1442, opened 2020-07-11
Bloc generator. Request for creating more detailed template of the bloc.


1g0rrr:
The generator is useful, but users need to write a lot of code on top of it. My suggestions:

1. Create basic states:
We only have 4 states in every bloc: Initial, Success, Failure, InProgress. Why not create it immediately? Or at least a couple of them.

```dart
class BlocNameInProgress extends BlocNameState {
  const BlocNameInProgress ();
}
```
2. Create a sample event. It will be easier to copy or change it, instead of writing from scratch:
```dart
class SampleActionHandled extends BlocNameEvent {
  const SampleActionHandled();
}
```

3. Create sample conversion from event to state. It will be easier to copy existing code and cause fewer errors in the conversion function. (I had problems with all this yields, in process of learning BLoC).
```dart
  @override
  Stream<BlocNameState> mapEventToState(
    BlocNameEvent event,
  ) async* {
    if (event is SampleActionHandled) {
      yield* _mapSampleActionHandledToState(event, state);
    }
  }

  Stream<BlocNameState> _mapSampleActionHandledToState(
    SampleActionHandled event,
    BlocNameState state,
  ) async* {
    yield BlocNameSuccess();
  }
```
4. Set Equatable checked by default or remove this checkbox.
BlocBuilder without Equatable just rebuilds everything every time. For a better understanding of the whole reactive programming concept, it's a good idea to learn equatable from the start. I know it could cause problems when people forget to override "props", but now as I can see, there is another problem, newbies just omit this checkbox.
Maybe we can describe Equatable and 'props' more in comments in code to prevent its misuse?

5. Create a folder for the bloc
It's useful to have a folder with the bloc name. Could be a checkbox in dialog.

LeonIPierre:
1. I have a code snippet that this based on the 5 product states (empty, partial, loading, ideal, error) below is an example output. I added a message at the top for new users to have a better understanding on how states are handled. @felangel do you think it would be better served to have the states in separate files via an extension?

```dart
///Basic application states. By default Bloc only updates when the incoming
///and previous state difer(see https://bloclibrary.dev/#/faqs?id=state-not-updating) so change the 'props' property
///to ensure the states are different
abstract class TestState extends Equatable {
@override
List<Object> get props => [];
}

class TestIdealState extends TestState {
  final TestResult result;
  TestIdealState(this.result);
@override
List<Object> get props => [result];
}

class TestEmptyState extends TestState {}

[...]

PR #1603, opened 2020-08-12
feat(vscode): update snippets and documentation

zepfietje:
## Status

**READY**

## Breaking Changes

NO

## Description

This pull request merges the snippets from https://github.com/zepfietje/vscode-bloc-snippets and https://github.com/zepfietje/vscode-cubit-snippets so we have all convenient snippets in the main VS Code extension.

Also, some handy basic state and event snippets have been added. Partially related to https://github.com/felangel/bloc/issues/1442.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/extensions/vscode/snippets/bloc.json
+++ b/extensions/vscode/snippets/bloc.json
@@ -1,4 +1,25 @@
 {
+  "Bloc": {
+    "prefix": "blocclass",
+    "body": [
+      "class ${1:Subject}Bloc extends Bloc<$1Event, $1State> {",
+      "\t${1:Subject}Bloc() : super(${1:Subject}Initial());",
+      "",
+      "\t@override",
+      "\tStream<$1State> mapEventToState($1Event event) async* {",
+      "\t\t${3:// TODO: implement mapEventToState}",
+      "\t}",
+      "}"
+    ]
+  },
+  "Cubit": {
+    "prefix": "cubitclass",
+    "body": [
+      "class ${1:Subject}Cubit extends Cubit<$1State> {",
+      "\t${1:Subject}Cubit() : super(${1:Subject}Initial());",
+      "}"
+    ]
+  },
   "BlocBuilder": {
     "prefix": "blocbuilder",
     "body": [
--- a/extensions/vscode/snippets/bloc.json
+++ b/extensions/vscode/snippets/bloc.json
@@ -1,25 +1,4 @@
 {
-  "Cubit": {
-    "prefix": "cubit",
-    "body": [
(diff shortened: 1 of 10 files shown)


extensions/vscode/snippets/bloc.json
```diff
@@ -1,25 +1,4 @@
 {
-  "Cubit": {
```
felangel, line 2: why do you want to remove these? I would argue they're still useful in situations where you just want to create a new bloc/cubit in the existing file

zepfietje: @felangel, mainly to use the `bloc` prefix for `context.bloc()`.

zepfietje: We could keep the bloc and cubit class snippets, but I'd suggest adding a different prefix.

extensions/vscode/snippets/bloc.json
```diff
@@ -74,96 +150,91 @@
...
-      "\t\tsuper.onEvent(bloc, event);",
-      "\t}",
+      "$4",
       "",
       "\t@override",
-      "\tvoid onError(Cubit cubit, Object error, StackTrace stacktrace) {",
-      "\t\t${3:// TODO: implement onError}",
-      "\t\tsuper.onError(cubit, error, stacktrace);",
-      "\t}",
+      "\tList<Object> get props => [$6];",
```
felangel, line 178: I don't think we should default to Equatable especially since there is technically also freezed support. Thoughts?

zepfietje: I think, more often than not, you do want to use Equatable. Especially when using the initial, in progress, success and failure states. This general state snippet is based on that.

What would you suggest?

felangel: Yeah thinking about it some more and I'm okay with defaulting to Equatable for now but I don't think we should include the initial, in progress, success, and failure states because we already have the `blocevent` and `blocstate` snippets and they are a bit too opinionated/specific imo. I also kept the `bloc` and `cubit` prefixes but renamed `contextbloc` to `ctxbloc` etc... so hopefully it's a good compromise because it's clear and more concise.

zepfietje: @felangel, the initial, in progress, success and failure states (as well as the started event) are entirely based on the naming conventions in the docs. Why do you think they're still too opinionated?

Also, I don't really like `ctxbloc`, since it's not something you're used to typing. I'd rather suggest something like `cbloc` or reverting to `contextbloc`. Thoughts?

PR #1634, opened 2020-08-17
refactor: BlocWidgetMixin and BlocWidgetStateMixin help to handle cubit subscription without code overhead

kranfix:
## Status

**READY**

## Breaking Changes

NO

## Description

`BlocListener` and `BlocBuilder` share code for subscription to a cubit and it could be encapsulated with mixins (`BlocWidgetMixin` and `BlocWidgetStateMixin`).
On the other hand, `BlocConsumer` is a `StatelessWidget` that compose a `BlocListener` with a `BlocBuilder`, then this widget makes a double effort for handle the same subscription, previous and current state.

With the proposed mixins, both `BlocListener` and `BlocBuilder` are simplified, while the implementation of `BlocConsumer` only extends a `BlocBuilder` (with some override methods), then there is only one subscription handling without many widget composition.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
Hi @kranfix 👋 
Thanks for taking the time to open a pull request, I really appreciate it! I'll take a closer look at this tomorrow and try to get a solution merged asap. I'm thinking it might be simpler to have an internal base class which handles the subscriptions and exposes the callback for onStateChange rather than using mixins but will play with a few approaches and get back to you as soon as possible! Thanks again 🙏

kranfix:
The `void onStateChanged(...)` was refactored to `bool onStateEmitted(...)` and if it returns true, the widget rebuilds.

felangel:
@kranfix I've opened #1679 to address this issue by taking advantage of BlocListener as the base for `BlocBuilder` and `BlocConsumer` this will allow `BlocConsumer` to only require a single subscription without introducing a lot of new concepts/components like `BlocWidgetStateMixin` and `BlocWidgetMixin `. Let me know if that addresses the issue and I think we can close this PR, thanks! 🙏

kranfix:
@felangel you solution is very simple :+1: . But what about the reuse of code. Both `BlocListener` and `BlocBuilder` state repeats each other to handle the `StreamSubscription`.  
  
That's why the mixins and they don't have to be exposed.

felangel:
@kranfix I've refactored that as well so `BlocBuilder` uses `BlocListener` and `BlocListener` is the only one which manages the `StreamSubscription`. Check out https://github.com/felangel/bloc/blob/a1305e56c74bb600025e1fe510a13a03bf14d106/packages/flutter_bloc/lib/src/bloc_builder.dart#L121 and let me know what you think 👍

kranfix:
Perfect! I say nothing.  
  
The most important thing is that the PR's goals were met:  
:white\_check\_mark:  Removing duplicated code  
:white\_check\_mark:  Optimizing BlocConsumer

Issue #1639, opened 2020-08-19
[bloc_test] when verify fails the exception does not bubble up in blocTest


nerder:
**Describe the bug**
When performing `verify` step in the `blocTest` in case of failure the exception is not bubbled up making the test fail, instead fails because of the timeout.

**To Reproduce**

Is possible to see the bug in this repo: https://github.com/nerder/bloc_test_verify_issue

**Expected behavior**
The expected behavior will be that the test fails as soon as the verify condition fails, instead of waiting for the timeout.

PR #1642, opened 2020-08-19
WIP: test should fail when verify throws an exception

nerder:
## Status

**IN DEVELOPMENT**

## Breaking Changes

NO

## Description

I've added the ability to bubble up `verify` errors so that is possible to debug when there are some unexpected failures.

Fixes https://github.com/felangel/bloc/issues/1639

## Help Needed

The tests are not passing, because I need to find a way to say that is ok that `blocTest` fails, I've tried wrapping it inside a `test` bloc, but it seems not to be possible, since `blocTest` is calling `test` internally. Therefore I've been given this error `StateError:<Bad state: Can't call test() once tests have begun running.>` which is not the one that I expect.

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -140,7 +140,7 @@ void blocTest<C extends Cubit<State>, State>(
         await subscription.cancel();
         try {
           await verify?.call(cubit);
-        } catch(error) {
+        } catch (error) {
           verifyError += error.toString();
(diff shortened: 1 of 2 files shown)


nerder:
I'm closing this for now since it might be possible that we should have a discussion about how to handle failed expectations in the assertive properties of `blocTest` such as `expect` and `verify`.

PR #1828, opened 2020-10-14
feat: added example code login using hydrated_bloc

dafinoer:
## Status

**READY**

## Breaking Changes

NO

## Description

im added some example code login app with hydrated bloc

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
Hi @dafinoer 👋 
Thanks so much for opening a pull request! What are your thoughts on just updating the existing `flutter_login` example to use `HydratedBloc`? I would rather do that than duplicate the login code for maintainability. Thoughts?

dafinoer:
i think that was a new scenario login with HydratedBloc, but it's okey, i will updating your flutter_login, i think it's a make sense.

felangel:
> i think that was a new scenario login with HydratedBloc, but it's okey, i will updating your flutter_login, i think it's a make sense.

Thanks! It becomes quite time consuming to maintain all of the examples so I'd really prefer to have a single example app that illustrates multiple concepts rather than an example app per concept. Sorry for any inconvenience!

felangel:
> it's okay @felangel 😄
> 
> if i do create new branch for updating flutter_bloc with hydrated_bloc, it's a good idea?

yeah we can close this pull request and open a new one to add hydrated_bloc to login or a different example. Might actually be good to add it to something like the infinite list example since it is more applicable. What do you think?

Issue #1980, opened 2020-11-30
BLoC for session events


enyo:
In the documentation it says:

> Blocs should never directly emit new states. Instead every state change must be output in response to an incoming event within mapEventToState.

I was under the impression, that BLoC should be used everywhere business logic should be abstracted from the UI. How would you suggest to handle session state specific information?

As an example: I have a nav bar, and want to write "Logged in as USERNAME | Logout" or "Sign In!" depending on user state. This state needs to change in response to the session state, so I would have provided the `SessionService` to a `SessionBloc` and used that to control the state by listening to changes on the `SessionService` and emitting new states in my BLoC.

Since that is discouraged, is there another way you would approach this?

felangel:
Hi @enyo 👋 
Thanks for opening an issue!

> Blocs should never directly emit new states.

The above statement refers to the fact that you shouldn't do something like:

```dart
RaisedButton(
  onPressed: () => myBloc.emit(newState),
);
```

You can still have the `SessionBloc` listen to state changes in the `SessionService` and add events internally 👍 
Hope that helps!

enyo:
Great. That's how I understood it. I think that the docs on the website are misleading

> Blocs should never directly emit new states. Instead every state change must be output in response to an incoming event within mapEventToState.

I think that it should be something like:

"New states should only be emitted by blocs themselves. Either as a reaction to an incoming event within `mapEventToState` or because an underlying state changed (e.g.: the user authenticated)."

enyo:
The `emit` function itself also has a `@visibleForTesting` annotation, and the docblock:

> **[emit] should never be used outside of tests.**
> Updates the state of the bloc to the provided [state].
> A bloc's state should only be updated by `yielding` a new `state`
> from `mapEventToState` in response to an event.

I think, it would make sense to change this to `@protected` and change the docblock to say that usually, it should only be used by `mapEventToState` or when responding to other state change events.

felangel:
closing this based on our conversation in #1982 👍

PR #1982, opened 2020-12-01
Remove @visibleForTesting annotation on emit

enyo:
## Status

**READY**

## Breaking Changes

NO

## Description

As discussed in #1980 . I think that `emit()` should only be `@protected` and not `@visibleForTesting` since there are cases where emitting a state without being a direct response of a user event is perfectly reasonable.

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/docs/coreconcepts.md
+++ b/docs/coreconcepts.md
@@ -334,7 +334,7 @@ In the above snippet, we are switching on the incoming event and if it is an inc
 
 ?> **Note**: Since the `Bloc` class extends `Cubit`, we have access to the current state of the bloc at any point in time via the `state` getter.
 
-!> Blocs should never directly `emit` new states. Instead every state change must be output in response to an incoming event within `mapEventToState`.
+!> Code that uses Blocs should never directly invoke `emit()` to emit new states. Instead every state change must be output in response to an incoming event within `mapEventToState` or in response to an internal state change (e.g.: the user signed in or out).
 
 !> Both blocs and cubits will ignore duplicate states. If we yield or emit `State nextState` where `state == nextState`, then no state change will occur.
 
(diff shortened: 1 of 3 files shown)


enyo:
Interesting... so there is a problem: adding `@visibleForTesting` triggers an error when trying to access the function outside of tests, but removing `@visibleForTesting` triggers an error in tests due to `@protected`. That seems like it's an issue with the meta library.

felangel:
@enyo I don't think this change is needed. I'm sorry if there was a miscommunication in #1980 but you should never emit from within a bloc. The `emit` API is reserved for `cubit` only and should only ever be used internally within the cubit.

enyo:
@felangel Ok, but how would you then handle a bloc, that changes it's state depending on events that are not triggered by the user? Let's say you have a menu bar, which shows the user information, and it needs to change to a logged out view when that changes, but it's not in charge of actually handling any sing in/out events by the user.

Normally, I'd simply provide some `SessionService` to the Bloc, and would listen to changes in the session to reflect the change.

How do you propose to do it instead?

felangel:
@enyo you can just add events internally

```dart
subscription = service.stream.listen((data) {
  add(DataChanged(data));
});
```

enyo:
@felangel Ok I see. Is there any advantage taking this detour of creating an event that is only used internally?

felangel:
It's not a detour -- all state changes in bloc must be triggered by an incoming event. Changes based on internal subscriptions are no exception. If you don't use an event and try to emit directly you will lose information about the state change (transition) and will bypass all processing events which can lead to unpredictable behavior.

enyo:
Thanks. That makes sense!

PR #2027, opened 2020-12-19
refactor(bloc)!: minor code simplification

wujek-srujek:
## Status

**READY**

## Breaking Changes

NO

## Description

Just a couple of minor code simplifications, please see inline comments in the PR.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -25,7 +25,7 @@ abstract class Bloc<Event, State> extends Cubit<State>
   /// The current [BlocObserver].
   static BlocObserver observer = BlocObserver();
 
-  final _eventController = StreamController<Event>.broadcast();
+  final _eventController = StreamController<Event>();
 
   StreamSubscription<Transition<Event, State>> _transitionSubscription;
 
--- a/packages/bloc/lib/src/cubit.dart
+++ b/packages/bloc/lib/src/cubit.dart
@@ -61,7 +61,7 @@ abstract class Cubit<State> extends Stream<State> {
 
   BlocObserver get _observer => Bloc.observer;
 
-  StreamController<State> _controller;
+  final StreamController<State> _controller = StreamController.broadcast();
 
   State _state;
 
@@ -79,7 +79,6 @@ abstract class Cubit<State> extends Stream<State> {
   @protected
   @visibleForTesting
   void emit(State state) {
-    _controller ??= StreamController<State>.broadcast();
     if (_controller.isClosed) return;
     if (state == _state && _emitted) return;
     onChange(Change<State>(currentState: this.state, nextState: state));
@@ -156,7 +155,6 @@ abstract class Cubit<State> extends Stream<State> {
     void Function() onDone,
     bool cancelOnError,
   }) {
-    _controller ??= StreamController<State>.broadcast();
     return _controller.stream.listen(
       onData,
       onError: onError,
@@ -176,7 +174,6 @@ abstract class Cubit<State> extends Stream<State> {
   Future<void> close() {
     // ignore: invalid_use_of_protected_member
     _observer.onClose(this);
-    _controller ??= StreamController<State>.broadcast();
     return _controller.close();
   }
 }


packages/bloc/lib/src/bloc.dart
```diff
@@ -25,7 +25,7 @@ abstract class Bloc<Event, State> extends Cubit<State>
   /// The current [BlocObserver].
   static BlocObserver observer = BlocObserver();
 
-  final _eventController = StreamController<Event>.broadcast();
```
wujek-srujek, line 28: The event `StreamController` is only ever subscribed to once, in `_bindEventsToStates ` which is called by the constructor, so there seems to be no need for using a broadcast stream.

packages/bloc/lib/src/cubit.dart
```diff
@@ -61,7 +61,7 @@ abstract class Cubit<State> extends Stream<State> {
 
   BlocObserver get _observer => Bloc.observer;
 
-  StreamController<State> _controller;
+  final StreamController<State> _controller = StreamController.broadcast();
```
wujek-srujek, line 64: The state `StreamController` is crated conditionally in 3 different places, including `close`. This means that even though it is lazily created, it will be created at least once, when the cubit is closed. Why not create it right away? It will also be created only once, but the advantages are:

- It can be `final`.
- Easier code - no need for `_controller ??= ...` in 3 different places. Especially the one in `close` seems weird, but I guess it has to do with the fact that you want the `done` event to be emitted on `close`?

packages/bloc/lib/src/bloc.dart
felangel: Great catch! Thanks for the improvement 👍

packages/bloc/lib/src/cubit.dart
felangel: I would prefer reverting this change because lazily creating the StreamController reduces the memory footprint of a bloc/cubit. While the controller will always be created, it will only be created when it is absolutely necessary instead of consuming unnecessary resources when it's not in use.

> but I guess it has to do with the fact that you want the done event to be emitted on close?
That's correct.

felangel (changes requested):
See my comment regarding reverting the `StreamController` changes and let me know what you think 👍

packages/bloc/lib/src/cubit.dart
wujek-srujek: I will do as you wish, but since you asked for my thoughts, here they are ;d

1. In Flutter, `BlocProvider` is lazy by default and created when the user first fetches it from the `BuildContext`. For the controller to be uninitialized when `Cubit.close` is called means that the dev did something like this: ```final b = context.read<SomeBloc>();``` and never, ever added any event, invoked a method changing the state or subscribed or whatever. This seems to be fishy code, likely a mistakte and I don't think the library needs to optimize for this case, code readability being the tradeoff.
2. In Flutter with an eager `BlocProvider`, the user needs to explicitly say they want it to be eager. If they do, eager should mean eager, so why would you make internal things lazy even in this case?
3. In Dart only, no Flutter - my arguments get weaker ;d but I think that if a user does ```final bloc = SomeBloc();``` and never uses it, it also means there is an issue in the code and there is no need to optimize in this case, similar to number 1 above.

Let me know and I will revert the `Cubit` changes.

packages/bloc/lib/src/cubit.dart
```diff
@@ -176,7 +174,6 @@ abstract class Cubit<State> extends Stream<State> {
   Future<void> close() {
     // ignore: invalid_use_of_protected_member
     _observer.onClose(this);
-    _controller ??= StreamController<State>.broadcast();
```
wujek-srujek, line 179: As a side question: why is it necessary for a not fully initialized `Cubit` (without the controller being set) to emit 'done' event? If it is uninitialized, it means nobody ever subscribed to it, so nobody cares whether 'done' is emitted or not, right? I know there are some tests for it but as per my understanding it looks like they are contrived, test something that would never occur outside of tests and in the real world nobody would wait on the 'done' event since nobody ever subscribed (otherwise, if anybody subscribes, the controller is initialized and this line in `close` will have no effect whatsoever and 'done' will be emitted).

What am I missing? What is the use case for emitting 'done' even without any subscribed being important?

PR #2069, opened 2021-01-11
infinite list does not get more records at the end: fixed

hansbak:
## Status

**READY

## Breaking Changes

YES 

## The infinite list example does not fetch new posts at the end of the screen

- [ x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/examples/flutter_infinite_list/lib/posts/bloc/post_bloc.dart
+++ b/examples/flutter_infinite_list/lib/posts/bloc/post_bloc.dart
@@ -77,5 +77,5 @@ class PostBloc extends Bloc<PostEvent, PostState> {
     throw Exception('error fetching posts');
   }
 
-  bool _hasReachedMax(int postsCount) => postsCount < _postLimit ? false : true;
+  bool _hasReachedMax(int postsCount) => postsCount < _postLimit ? true : false;
 }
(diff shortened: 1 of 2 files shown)


examples/flutter_infinite_list/lib/posts/bloc/post_bloc.dart
```diff
@@ -77,5 +77,5 @@ class PostBloc extends Bloc<PostEvent, PostState> {
     throw Exception('error fetching posts');
   }
 
-  bool _hasReachedMax(int postsCount) => postsCount < _postLimit ? false : true;
+  bool _hasReachedMax(int postsCount) => postsCount < _postLimit ? true : false;
```
felangel, line 80: The old logic seems correct to me. If the postsCount is less than the limit, we have not reached the max. The updated implementation seems incorrect.

For example, if `postsCount` is 0 and `_postLimit` is 20, with the proposed changes `_hasReachedMax` would evaluate to `true` which is incorrect.

felangel (changes requested):
See comments, I don't think these changes are correct. Can you please provide more information regarding why you feel there is a bug? I am not able to reproduce the issue described.

felangel:
Closing for now based on my previous comment

PR #2165, opened 2021-02-18
ci(bloc_test): update bloc_test.yaml workflow with new dart syntax

PlugFox:
Update bloc_test.yaml workflow with new dart syntax as example

### DIFF
--- a/.github/workflows/bloc_test.yaml
+++ b/.github/workflows/bloc_test.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Install Dependencies
         run: pub get
       - name: Format
-        run: dart format --set-exit-if-changed none
+        run: dart format --set-exit-if-changed .
       - name: Analyze
         run: dart analyze --fatal-infos --fatal-warnings .
       - name: Run tests
--- a/.github/workflows/bloc_test.yaml
+++ b/.github/workflows/bloc_test.yaml
@@ -24,11 +24,11 @@ jobs:
       - name: Install Dependencies
         run: pub get
       - name: Format
-        run: dartfmt --dry-run --set-exit-if-changed .
+        run: dart format --set-exit-if-changed none
       - name: Analyze
         run: dart analyze --fatal-infos --fatal-warnings .
       - name: Run tests
-        run: dart test --coverage=coverage && pub run coverage:format_coverage --lcov --in=coverage --out=coverage/lcov.info --packages=.packages --report-on=lib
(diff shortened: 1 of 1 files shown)


.github/workflows/bloc_test.yaml
```diff
@@ -24,11 +24,11 @@ jobs:
       - name: Install Dependencies
         run: pub get
       - name: Format
-        run: dartfmt --dry-run --set-exit-if-changed .
+        run: dart format --set-exit-if-changed none
```
felangel, line 27: I believe `dart format` expects a directory so in this case the command does not format any files because a directory called `none` does not exist
```suggestion
        run: dart format --set-exit-if-changed .
```

Screen Shot 2021-02-18 at 3 50 43 PM

felangel (changes requested):
Thanks for the pull request! See my comment and let me know what you think 👍

.github/workflows/bloc_test.yaml
PlugFox: Nope)
One second.

PlugFox:
```bash
PS > dart format -h
Idiomatically format Dart source code.

Usage: dart format [options...] <files or directories...>
-h, --help                     Print this usage information.
-v, --verbose                  Show all options and flags with --help.
-o, --output                   Set where to write formatted output.

          [json]               Print code and selection as JSON.
          [none]               Discard output.
          [show]               Print code to terminal.
          [write] (default)    Overwrite formatted files on disk.

    --set-exit-if-changed      Return exit code 1 if there are any formatting changes.
    --fix                      Apply all style fixes.
-l, --line-length              Wrap lines longer than this.
                               (defaults to "80")

Run "dart help" to see global options.
```

You anyway can't change files under git.

With option
`[none] - Discard output.`
we only check format source code and throw error if something wrong.

PlugFox: Also, syntax is changed.
If you anyway want `.`, probably you want `write` instead.

PlugFox: @felangel sorry, my bad.  Must be: `dart format --set-exit-if-changed -o none .`

PlugFox:
@felangel sorry, my bad. Must be: `dart format --set-exit-if-changed -l 80 -o none .`

felangel (approved):
Thanks for the contributions! 💯 🚀

PlugFox:
@felangel i may change other workflows with same changes

felangel:
> @felangel i may change other workflows with same changes.
> What do you think?

That would be great, thanks!

PR #2258, opened 2021-03-21
docs: fixed imports

pythondev0101:
Not to confuse new developers on using "part"

## Status

**READY**

## Breaking Changes

NO

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
Hi @pythondev0101 👋 

Thanks for opening a PR but I’d prefer to leave it as is because this is how we recommend structuring blocs and it is also how the blocs will be generated when using the vscode or IntelliJ extensions. If there is confusion about parts I’d prefer to explain them rather than modifying the code. Thanks! 🙏

PR #2297, opened 2021-04-04
Example/flutter notification example

AlankritNayak:
## Status

**IN DEVELOPMENT**

## Breaking Changes

NO

## Description

Added an example flutter app, which implements Push Notification, using Firebase Cloud Message service (FCM).  
- The app is capable of showing push notification while its in foreground or background or terminated
- It has a NotificationRepository, which handles showing notification and exposing a stream of notifications, which the NotificationBloc can subscribe to.
- NotificatinoBloc can then update UI based on the data in this stream
- ### Remaining features to be implemented: 
    - The NotificationRepository currently exposes a stream which contains only one notification, it should be modified so that it emits a list of latest notifications
    - When the app launches as a result of user tapping a notification, it should directly open Notifications page, instead of home screen. 
    - When app is in terminated/background state, and a notification is received, we should save this notification in local storage, and retrieve them on next startup.

- ### More details are in the REAMDE.md file

## Type of Change

- [x] 📝 Example


examples/flutter_notifications/lib/home.dart
```diff
@@ -0,0 +1,25 @@
...
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text('Home Page'),
+        actions: [
+          IconButton(
+              icon: Icon(Icons.notifications),
+              onPressed: () {
+                Navigator.pushNamed(context,NotificationPage.routeName);
+              })
```
marcossevilla, line 17:
```suggestion
              },
            ),
```

examples/flutter_notifications/lib/main.dart
```diff
@@ -0,0 +1,61 @@
...
+}
+
+class _MyAppState extends State<MyApp> {
+  final _navigatorKey = GlobalKey<NavigatorState>();
+
+  NavigatorState? get _navigator => _navigatorKey.currentState;
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocProvider<NotificationBloc>(
```
marcossevilla, line 35: Can we move the dependency injection to the App widget and consume the bloc you create at an AppView widget? That way, we avoid using the builder property from `MaterialApp`.

You can check this DI pattern in [this](https://github.com/felangel/bloc/blob/master/examples/flutter_weather/lib/weather/view/weather_page.dart) example with `WeatherPage` and `WeatherView`.

examples/flutter_notifications/lib/notification/bloc/notification_bloc.dart
```diff
@@ -0,0 +1,48 @@
...
+    _notificationSubscription = _notificationRepository.notificationStream
+        .listen((NotificationEntity notificationEntity) {
+      add(NotificationReceived(notificationEntity));
+    });
+  }
+
+  final NotificationRepository _notificationRepository;
+  late StreamSubscription<NotificationEntity> _notificationSubscription;
+  @override
+  Stream<NotificationState> mapEventToState(
```
marcossevilla, line 27: As of v7.2.0, `mapEventToState` is deprecated. In order to merge this PR, you'll have to migrate to v7.2.0 or higher. You can check the migration guide [here](https://bloclibrary.dev/#/migration?id=v720).

examples/flutter_notifications/lib/notification/view/widgets/notification_list.dart
```diff
@@ -0,0 +1,28 @@
...
+    Key? key,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocBuilder<NotificationBloc, NotificationState>(
+      builder: (context, state) {
+        if (state is NotificationRecieveSuccess) {
+          return ListView.builder(
+              itemCount: 1,
```
marcossevilla, line 16: Are we only showing 1 notification at a time?

examples/flutter_notifications/notification_repository/lib/src/notification_repository_fcm.dart
```diff
@@ -0,0 +1,146 @@
...
+    _streamController.close();
+  }
+
+  @override
+  Future<void> initialize() async {
+    await Firebase.initializeApp();
+    _flutterLocalNotificationsPlugin = FlutterLocalNotificationsPlugin();
+    _firebaseMessaging = FirebaseMessaging.instance;
+    String? token = await FirebaseMessaging.instance.getToken();
+    print(token);
```
marcossevilla, line 56: We should avoid having print calls on production code.

examples/flutter_notifications/notification_repository/lib/src/notification_repository_fcm.dart
```diff
@@ -0,0 +1,146 @@
...
+    _setupOnMessageListener();
+    FirebaseMessaging.onBackgroundMessage(firebaseMessagingBackgroundHandler);
+    await _handleInteraction();
+    if (Platform.isAndroid)
+      await _enableAndroidForegroundNotification();
```
marcossevilla, line 61: Can you add curly brackets to this if-else?

examples/flutter_notifications/notification_repository/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
+name: notification_repository
+description: repository to handle displaying notificaiton and exposing a stream of notificaitons
```
marcossevilla, line 2:
```suggestion
description: A repository to handle displaying and exposing a stream of notifications.
```

marcossevilla (changes requested):

examples/flutter_notifications/lib/notification/view/widgets/notification_list.dart
AlankritNayak: Yes I initially only implemented 1 notification. multiple notifications can be implemented in future.

PR #2312, opened 2021-04-08
Exporting mocktail from bloc_test

albertodev01:
## Status

**READY**

## Breaking Changes

NO

## Description

I have simply added the `export 'package:mocktail/mocktail.dart';` in the `bloc_test` package because when writing tests for a bloc, you need (for example) to call `registerFallbackValue` (which is part of mocktail). In this way:

  - you can only use `import 'package:bloc_test/bloc_test.dart';` and Dart will automatically import mocktail too
  - the developer has only 1 dependency to update (bloc_test) rather than 2 (bloc_test AND mocktail)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/packages/bloc_test/lib/bloc_test.dart
+++ b/packages/bloc_test/lib/bloc_test.dart
@@ -1,5 +1,7 @@
 library bloc_test;
 
+export 'package:mocktail/mocktail.dart';
+
 export 'src/bloc_test.dart';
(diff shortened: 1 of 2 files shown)


felangel (approved):
LGTM! Thanks so much for the contribution! 💯 🚀 💙

PR #2463, opened 2021-05-20
fix(extensions): Update override props method to have return type List of nullable objects.

aadarshadhakalg:
## Status

**READY**

## Breaking Changes

NO

## Description

 The snippet/templates of bloc events and bloc state has the method overriding props with return type List<Object>. I change it to List<Object?> because the props can be null sometimes. 

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/extensions/intellij/intellij_generator_plugin/src/main/resources/liveTemplates/Bloc.xml
+++ b/extensions/intellij/intellij_generator_plugin/src/main/resources/liveTemplates/Bloc.xml
@@ -87,7 +87,7 @@
         </context>
     </template>
     <template name="blocevent"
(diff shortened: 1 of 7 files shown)


felangel:
Hi @aadarshadhakalg 👋 
Thanks for opening a pull request!

I would prefer not to make props nullable by default because it's unnecessary. I would much rather have developers make them nullable only if needed because it's a simple change. I recommend not making props nullable as just as a preventative measure. Closing for know but I'm happy to continue the conversation if you feel strongly about this change. Thanks! 🙏

PR #2469, opened 2021-05-23
fix(bloc_test): allow blocTest to capture throw non-errors

cameron1024:


felangel (approved):
LGTM thanks so much for the contribution! 💯 💙 🚀

PR #2181, opened 2021-03-02
fix(flutter_bloc): BlocBuilder updates when provided bloc instance changes

felangel:
## Status

**IN DEVELOPMENT**

## Breaking Changes

NO

## Description

- fix(flutter_bloc): BlocBuilder updates when provided bloc instance changes (closes #2127)
  - related to #2124

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [X] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #2482, opened 2021-05-27
fix(flutter_bloc): Add dependency to bloc instance for BlocBuilder/BlocListener/BlocConsumer

### LINKED ISSUE #2127: `BlocBuilder` continues to use old cubit instance when the context is updated with a new instance
**Describe the bug**
When injecting a cubit with `BlocProvider.value` and then building with `BlocBuilder`, changing the cubit to a new instance does not cause the `BlocBuilder` to subscribe to state changes in the new cubit.

**To Reproduce**

My desired use case is to replace the bloc instance in the context when some input changes, automatically closing the old instance:
```
/// A widget that acts identically to [BlocProvider] except it will close its
/// cubit and recreate it when [input] changes.
class BlocInputProvider<C extends Cubit, Input> extends StatefulWidget {
  final Input input;
  final C Function(BuildContext, Input) create;
  final Widget child;

  BlocInputProvider({
    Key key,
    @required this.input,
    @required this.create,
    @required this.child,
  }) : assert(child != null);

  @override
  _BlocInputProviderState<C, Input> createState() =>
      _BlocInputProviderState<C, Input>();
}

class _BlocInputProviderState<C extends Cubit, Input>
    extends State<BlocInputProvider<C, Input>> {
  C _cubit;

  @override
  void initState() {
    super.initState();

    _createCubit();
  }

  @override
  void dispose() {
    super.dispose();

    _cubit.close();
  }

  @override
  void didUpdateWidget(covariant BlocInputProvider<C, Input> oldWidget) {
    super.didUpdateWidget(oldWidget);

    if (widget.input != oldWidget.input) {
      _cubit.close();

      _createCubit();
    }
  }

  @override
    return BlocProvider<C>.value(
      value: _cubit,
      child: widget.child,
    );
  }

  void _createCubit() {
    _cubit = widget.create(context, widget.input);
  }
}
```

However, `BlocBuilder` in the subtree doesn't react to the cubit being replaced. Digging into `BlocBuilderBase`, it looks like it only checks for a new instance of the bloc in `didUpdateWidget` using `context.read`, meaning it won't know about the new cubit instance until its configuration changes. 

I'm reporting this as a bug, but it's very possible this is the intended behavior and there's a very good reason for it!

pr0gramista:
## Status

**READY**

## Breaking Changes

NO

## Description
Fixes #2127
Closes https://github.com/felangel/bloc/pull/2181

To make a builder, listener or consumer update when the instance is changed we need a context dependency. However we can't use normal `context.watch` because BlocProvider uses a custom provider which also updates dependents when state is changed, therefore it would make `buildWhen` useless. Ideally we would reverse that, but that would also break compatibility. Adding another "normal" provider is also not an option since provider does not allow it and integrating special InheritedWidget is tricky.

There is a concept of "aspect" when it comes to dependencies. Provider exposes it by `context.select`, but it is a bit different   and it can't be used in `didUpdateDependencies`, but we can put it in build and use it to target only the changes of the instance by retrieving hashCode.

I think for people who doesn't change instances in the runtime this makes almost no difference apart from dependency.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_builder.dart
+++ b/packages/flutter_bloc/lib/src/bloc_builder.dart
@@ -135,10 +135,9 @@ class _BlocBuilderBaseState<B extends BlocBase<S>, S>
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-
-    final newBloc = widget.bloc ?? context.read<B>();
-    if (_bloc != newBloc) {
-      _bloc = newBloc;
+    final bloc = widget.bloc ?? context.read<B>();
+    if (_bloc != bloc) {
+      _bloc = bloc;
       _state = _bloc.state;
     }
   }
@@ -156,10 +155,7 @@ class _BlocBuilderBaseState<B extends BlocBase<S>, S>
 
   @override
   Widget build(BuildContext context) {
-    if (widget.bloc == null) {
-      context.select<B, int>((bloc) => bloc.hashCode);
-    }
-
+    if (widget.bloc == null) context.select<B, int>(identityHashCode);
     return BlocListener<B, S>(
       bloc: _bloc,
       listenWhen: widget.buildWhen,
--- a/packages/flutter_bloc/lib/src/bloc_builder.dart
+++ b/packages/flutter_bloc/lib/src/bloc_builder.dart
@@ -136,7 +136,7 @@ class _BlocBuilderBaseState<B extends BlocBase<S>, S>
   void didChangeDependencies() {
     super.didChangeDependencies();
 
-    final newBloc = widget.bloc ?? context.watch<B>();
+    final newBloc = widget.bloc ?? context.read<B>();
     if (_bloc != newBloc) {
       _bloc = newBloc;
       _state = _bloc.state;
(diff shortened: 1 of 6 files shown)


pr0gramista:
@felangel seems like `very_good_analysis` got an update yesterday and this one example doesn't pass CI anymore

packages/flutter_bloc/lib/src/bloc_builder.dart
```diff
@@ -145,6 +155,7 @@ class _BlocBuilderBaseState<B extends BlocBase<S>, S>
 
   @override
   Widget build(BuildContext context) {
+    if (widget.bloc == null) context.select<B, int>(identityHashCode);
```
felangel, line 158: It's a very creative idea to solve the previous issue! I made one minor adjustment to use `identityHashCode` in case developers override `hashCode` manually.

felangel (approved):
Thanks so much for taking the time to address this issue! I really appreciate the creative solution and will try to get this merged shortly 💙 🙏

felangel:
> @felangel seems like `very_good_analysis` got an update yesterday and this one example doesn't pass CI anymore

Yup #2483 should resolve this 👍

Issue #2086, opened 2021-01-15
replay_bloc should allow a filter function to ignore some states


JohnGalt1717:
**Is your feature request related to a problem? Please describe.**
In the state queue of most pages there is typically Loading, Loaded, error etc.  Unfortunately, right now the ones that don't matter to reply are stored and it doesn't appear to be any way to intercept and filter those out before they get dumped in to the change tracker.

**Describe the solution you'd like**
I'd like a method that can be overridden that is simply defined as bool shouldReplay(state) => true;

This would allow us to filter out stuff.

**Describe alternatives you've considered**
Simply overriding emit doesn't work because the change stack is private and even if it did it would be less than ideal

evandrmb:
I don't know if this should be another issue, but I think would be nice to have a `undoTo` that works like `Navigator.popUntil`. That way we would go back directly to the specified State.

PR #2488, opened 2021-05-28
feat(replay_bloc): filter which states are saved to the undo stack

Kirpal:
## Status

**READY**

## Breaking Changes

NO

## Description
I added a `shouldReplay` method to `ReplayBloc` and `ReplayCubit`. The intention of this is described more in #2086, but basically, the idea is that a developer can override this method to filter out which states are saved in the undo/redo stack.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/replay_bloc/lib/src/replay_bloc.dart
+++ b/packages/replay_bloc/lib/src/replay_bloc.dart
@@ -8,13 +8,13 @@ abstract class ReplayEvent {
   const ReplayEvent();
 }
 
-/// Notifies a [ReplayBloc] of a Redo
+/// Notifies a [ReplayBloc] of a Redo.
 class _Redo extends ReplayEvent {
   @override
   String toString() => 'Redo';
 }
 
-/// Notifies a [ReplayBloc] of an Undo
+/// Notifies a [ReplayBloc] of an Undo.
 class _Undo extends ReplayEvent {
   @override
   String toString() => 'Undo';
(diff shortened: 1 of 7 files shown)


packages/replay_bloc/lib/src/replay_bloc.dart
```diff
@@ -102,31 +102,33 @@ mixin ReplayBlocMixin<Event extends ReplayEvent, State> on Bloc<Event, State> {
...
-        onTransition(Transition(
-          currentState: this.state,
-          event: event,
-          nextState: val,
-        ));
-        // ignore: invalid_use_of_visible_for_testing_member
-        super.emit(val);
-      },
-    ));
+    if (shouldReplay(this.state)) {
```
felangel, line 105: What if `shouldReplay` evaluates to false when the state changes occurs but at a later time evaluates to true for the same state? As a developer, I would expect that the `ReplayBloc` maintain all states regardless and "skip" certain states that don't meet my criteria rather than permanently excluding states from the change stack. I can try to put together an example of what I'm describing in a bit.

Kirpal: Alright, I refactored this to keep all states in the stack and perform the check when you try to `undo`. This way:
1. If there are no states where `shouldReplay` returns true, `canUndo` is false
2. If the latest state doesn't replay, it's popped and we keep trying until one does

felangel (changes requested):
I added two more test cases for `redo` because I think the current behavior is not desirable. If `shouldReplay` is overridden and `undo` is called, then when `redo` is called, I would expect the states which were filtered out by `shouldReplay` should not be emitted. Let me know what you think.

packages/replay_bloc/lib/src/replay_bloc.dart
felangel: That sounds good, now I think we're not handling `redo` properly (see my most recent comments).

felangel (approved):
LGTM ✅ 
Thanks so much for taking the time to add this improvement 💙🚀

PR #2527, opened 2021-06-14
docs(bloc): update `onChange` description in README

chinkysight:
## Status

**READY**

## Breaking Changes

NO

## Description

I have updated this only paragraph which was confusing the users about `onChange` method: 
>State changes in cubit begin with predefined function calls which can use the emit method to output new states. onChange is called on each state change and contains the current and next state.

To
>State changes in cubit begin with predefined function calls which can use the emit method to output new states. **onChange is called right before a state change occurs and contains the current and next state.**
## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore


felangel (approved):
LGTM, thanks for the contribution! 💯 💙

PR #2585, opened 2021-07-02
fix: sign up cubit password confirmation order

### LINKED ISSUE #2179: Signup Cubit issue with password confirmation order
**Describe the bug**
There is a bug on the signup screen (SignUp Cubit) if the user enters the password confirmation first and after that the password.
Even if both passwords match exactly, the SignUp Button will stay disabled. You then have to erase a character from the password field and re-type.

**To Reproduce**
Go to the sign up screen, and type a 6 character (or more) password in the ConfirmPasswordInput field, then repeat the same password in the PasswordInput field => The SignUp Button will still be disabled

**Expected behavior**
The order in which the password and confirmed password are typed should not matter. 

**Solution**
In the SignUp Cubit, change the passwordChanged function. The "status" parameter should take the variable confirmedPassword (instead of currently state.confirmedPassword)

duan-daniel:
## Status

**READY**

## Breaking Changes

NO

## Description

* Updates `SignUp Cubit` so that the `SignUp Button` will be enabled if both passwords match, regardless of the order in which the passwords were inputted. 
* closes #2179 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/examples/flutter_firebase_login/lib/sign_up/cubit/sign_up_cubit.dart
+++ b/examples/flutter_firebase_login/lib/sign_up/cubit/sign_up_cubit.dart
@@ -35,7 +35,7 @@ class SignUpCubit extends Cubit<SignUpState> {
       status: Formz.validate([
         state.email,
         password,
-        state.confirmedPassword,
+        confirmedPassword,
       ]),
     ));
   }
--- a/examples/flutter_firebase_login/test/sign_up/cubit/sign_up_cubit_test.dart
+++ b/examples/flutter_firebase_login/test/sign_up/cubit/sign_up_cubit_test.dart
@@ -115,6 +115,32 @@ void main() {
           ),
         ],
       );
+
+      blocTest<SignUpCubit, SignUpState>(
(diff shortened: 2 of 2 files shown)


examples/flutter_firebase_login/lib/sign_up/cubit/sign_up_cubit.dart
```diff
@@ -35,7 +35,7 @@ class SignUpCubit extends Cubit<SignUpState> {
       status: Formz.validate([
         state.email,
         password,
-        state.confirmedPassword,
+        confirmedPassword,
```
felangel, line 38: Nice find! Can we add a unit test to ensure this bug I’d never reintroduced?

felangel (changes requested):
Great job! Can we just add a unit test to prevent a regression?

felangel (approved):
LGTM thanks so much for the contribution! 🚀

PR #2620, opened 2021-07-25
feat(bloc_test): add `setUp` and `tearDown`

felangel:
## Status

**READY**

## Breaking Changes

NO

## Description

- feat(bloc_test): add `setUp` and `tearDown` methods

As a developer, I want to have the ability to execute set up and tear down code as part of a `blocTest` so that I don't have to clutter the `build` and `verify` api:

```dart
blocTest(
  setUp: () {
    // Execute any set up code needed prior to building the bloc.
    when(() => repository.someMethod()).thenReturn(...);
  },
  build: () => MyBloc(repository),
  act: (bloc) => bloc.add(...),
  expect: () => [...],
  tearDown: () {
    // Execute any tear down code needed after the test has run.
    verifyNoMoreInteractions(repository);
  },
);
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc_test/test/bloc_bloc_test_test.dart
+++ b/packages/bloc_test/test/bloc_bloc_test_test.dart
@@ -506,6 +506,17 @@ void main() {
         },
       );
 
+      blocTest<SideEffectCounterBloc, int>(
+        'custom setUp is executed before build/act',
+        setUp: () {
+          when(() => repository.sideEffect()).thenThrow(Exception());
+        },
+        build: () => SideEffectCounterBloc(repository),
+        act: (bloc) => bloc.add(CounterEvent.increment),
+        expect: () => const <int>[],
+        errors: () => [isException],
+      );
+
       test('fails immediately when verify is incorrect', () async {
         const expectedError =
             '''Expected: <2>\n  Actual: <1>\nUnexpected number of calls\n''';
--- a/packages/bloc_test/test/bloc_bloc_test_test.dart
+++ b/packages/bloc_test/test/bloc_bloc_test_test.dart
@@ -507,7 +507,7 @@ void main() {
       );
 
       blocTest<SideEffectCounterBloc, int>(
-        'custom setUp is executed before build/act',
+        'setUp is executed before build/act',
         setUp: () {
           when(() => repository.sideEffect()).thenThrow(Exception());
         },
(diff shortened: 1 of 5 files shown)


packages/bloc_test/test/bloc_bloc_test_test.dart
```diff
@@ -550,4 +561,30 @@ Alternatively, consider using Matchers in the expect of the blocTest rather than
...
+    tearDown(() {
+      expect(tearDownCallCount, equals(1));
+    });
+
+    blocTest<CounterBloc, int>(
+      'is called after the test is run',
+      build: () => CounterBloc(),
+      act: (bloc) => bloc.add(CounterEvent.increment),
+      verify: (bloc) {
+        state = bloc.state;
```
korzonkiee, line 582: Is it necessary to set the `state` variable here and then expect that it's correct in `blocTest.tearDown`? Maybe I'm missing something, but it doesn't seem related to testing the `tearDown` callback.

felangel: the goal of this was to ensure that tearDown executes last. This isn't how you'd use `tearDown` normally -- it's more to ensure that tearDown is called last.

korzonkiee: ok, that makes sense - thanks for claryfing.

korzonkiee (approved):
LGTM 👍

jorgecoca (approved):
I understand why these changes are suggested and its benefits, but I question I can see being raised is why/when to use these new two APIs under `blocTest`, and when to use the `setUp` and `tearDown` methods from the `test` package.

Should we add more documentation for that?

felangel:
> I understand why these changes are suggested and its benefits, but I question I can see being raised is why/when to use these new two APIs under `blocTest`, and when to use the `setUp` and `tearDown` methods from the `test` package.
> 
> Should we add more documentation for that?

In my opinion the setUp and tearDown in the group is for common setUp and tearDown code whereas the setUp and tearDown apis in blocTest are intended to be test-specific (e.g. stubbing repository responses, etc.). These are usually test-specific and currently developers have to include that logic in the `build` api which clutters it.

PR #2621, opened 2021-07-25
feat(flutter_bloc): BlocSelector

### LINKED ISSUE #1989: select feature is good but just for context bloc ! local bloc can not suppot select feature
hello

on new versions you added new features :    **read   watch  select**
these features are very good features and their vacancy was seen for a long time

see this watch select example please
`final isPositive = context.select((CounterBloc b) => b.state >= 0);
`

this example get CounterBloc from context and we can creating different scenarios for different UI very faster
it is excellent. i love it.
```
   Color myColor = context.select((CounterBloc b) {
      if(b.state == 0)
           return Colors.red;
      else if(b.state == 1)
           return Colors.blue;
      
      retuen Colors.green;
    });
```

some times our Bloc not in the context !!!
```
for example we have one local variable of Bloc

// we can access to state of this local bloc like this => counterBloc.state
class MyWidget extends StatelessWidget {

CounterBloc counterBloc = CounterBloc();

 @override

   
    return Text('counterBloc.state');
  }

}
```

**Problem is here**
**but we need creating one complex scenario with different state of our local bloc**
but this solution just for Blocs on the context !! 
we have not this good feature for local instance our Bloc !!
```
class MyWidget extends StatelessWidget {

CounterBloc counterBloc = CounterBloc();

  // here we see local bloc state like observer and observable
   Color myColor = counterBloc.state.select((CounterBloc b) {
      if(b.state == 0)
           return Colors.red;
      else if(b.state == 1)
           return Colors.blue;
      
      retuen Colors.green;
    });

 @override
    return Container(color : myColor);
  }

}

```

are you have solution for my problem ?
thank you very much

felangel:
## Status

**READY**

## Breaking Changes

NO

## Description

- feat(flutter_bloc): BlocSelector (closes #1989)

As a developer, I want to be able to select part of a bloc state and selectively rebuild UI in response to when the selected value changes. Currently this can be accomplishing via the `context.select` extension but it is not supported as a built-in widget (specifically in cases where the bloc is not available in the current `BuildContext`).

### Documentation

**BlocSelector** is a Flutter widget which is analogous to `BlocBuilder` but allows developers to filter updates by selecting a new value based on the current bloc state. Unnecessary builds are prevented if the selected value does not change. The selected value must be immutable in order for `BlocSelector` to accurately determine whether `builder` should be called again.

If the `bloc` parameter is omitted, `BlocSelector` will automatically perform a lookup using `BlocProvider` and the current `BuildContext`.

```dart
BlocSelector<BlocA, BlocAState, SelectedState>(
  selector: (state) {
    // return selected state based on the provided state.
  },
  builder: (context, state) {
    // return widget here based on the selected state.
  },
)
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/README.md
+++ b/packages/flutter_bloc/README.md
@@ -145,6 +145,23 @@ BlocBuilder<BlocA, BlocAState>(
 )
 ```
 
+### BlocSelector
+
+**BlocSelector** is a Flutter widget which is analogous to `BlocBuilder` but allows developers to filter updates by selecting a new value based on the current bloc state. Unnecessary builds are prevented if the selected value does not change. The selected value must be immutable in order for `BlocSelector` to accurately determine whether `builder` should be called again.
+
+If the `bloc` parameter is omitted, `BlocSelector` will automatically perform a lookup using `BlocProvider` and the current `BuildContext`.
+
+```dart
+BlocSelector<BlocA, BlocAState, SelectedState>(
+  selector: (state) {
+    // return selected state based on the provided state.
+  },
+  builder: (context, state) {
+    // return widget here based on the selected state.
+  },
+)
+```
+
 ### BlocProvider
 
 **BlocProvider** is a Flutter widget which provides a bloc to its children via `BlocProvider.of<T>(context)`. It is used as a dependency injection (DI) widget so that a single instance of a bloc can be provided to multiple widgets within a subtree.
@@ -205,7 +222,7 @@ In addition, `context.select` can be used to retrieve part of a state and react 
 final isPositive = context.select((CounterBloc b) => b.state >= 0);
 ```
 
-The snippet above will only rebuild if the state of the `CounterBloc` changes from positive to negative or vice versa.
+The snippet above will only rebuild if the state of the `CounterBloc` changes from positive to negative or vice versa and is functionally identical to using a `BlocSelector`.
 
 ### MultiBlocProvider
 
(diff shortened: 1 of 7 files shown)


Kirpal:
What's the reason for using a `BlocSelector` instead of a `BlocBuilder` with `buildWhen`?

jeroen-meijer (approved):
Great work 😄
LGTM :shipit:

jeroen-meijer:
> What's the reason for using a `BlocSelector` instead of a `BlocBuilder` with `buildWhen`?

I think `BlocSelector` can be implemented using a `BlocBuilder` with `buildWhen` (i.e., `BlocSelector` is a subset of `BlocBuilder` if you will), but they can be used for different things. I think `BlocSelector` is more clear in its goal than `BlocBuilder` is when it's used in this way, and it saves a bit of code, too.

```dart
// Instead of
BlocBuilder(
  buildWhen: (previous, current) => previous.foo.bar != current.foo.bar,
  builder: (context, state) {
    final value = state.foo.bar;
  },
);

// This way is shorter
BlocSelector(
  // Don't need to do a comparison
  selector: (state) => state.foo.bar,
  builder: (context, bar) {
    // We have access to `bar` immediately.
  },
);

**Comment by felangel** [comment]
> What's the reason for using a `BlocSelector` instead of a `BlocBuilder` with `buildWhen`?

`BlocSelector` would mainly be for cases where you want to rebuild with a subset of the state when . For example:

```dart
BlocSelector<LoginBloc, LoginState>(
  selector: (state) => state.email,
  builder: (context, email) {
    // Build UI based on `email` property changes
    // We only have access to email here.
  },
);
```

Whereas `buildWhen` with `BlocBuilder` is more of an optimization and not really intended to filter state for a sub-tree (even though you can technically accomplish it as @jeroen-meijer mentioned).

jorgecoca (approved):

PR #2700, opened 2021-08-27
fix: Update authentication_state.dart for null safety

FrancescoPenasa:
If user model contain fields that cannot be const or final as for example another model such as a profile of the user, this fix make the code work

## Status

**READY**

## Breaking Changes

 NO

## Description

changed user definition to accept non-constant values as an empty list

## Type of Change

- [ x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel (changes requested):
This is intentional in order to avoid having a nullable user and the example has already been migrated to support sound null safety -- closing for now since this is working as intended.

PR #2729, opened 2021-09-09
updated the select snippet

MehadeHassan:
## Status

**READY**

## Breaking Changes

NO

## Description

The new `select` snippet looks familiar to `read` and `watch` snippets. 

**Old `select` snippet:** 
`context.select((AuthenticationBloc bloc) => bloc.state)`

**New `select` snippet :**
` context.select<AuthenticationBloc, AuthenticationState>(
              (bloc) => bloc.state)` 

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/extensions/vscode/snippets/flutter_bloc.json
+++ b/extensions/vscode/snippets/flutter_bloc.json
@@ -128,13 +128,13 @@
     "prefix": "read",
     "body": "context.read<${1:Subject}${2|Bloc,Cubit,Repository|}>()"
   },
-  "context.select()": {
-    "prefix": "select",
-    "body": "context.select((${1:Subject}${2|Bloc,Cubit|} ${3:element}) => $3$4)"
-  },
   "context.watch()": {
     "prefix": "watch",
     "body": "context.watch<${1:Subject}${2|Bloc,Cubit,Repository|}>()"
+  },
+  "context.select()": {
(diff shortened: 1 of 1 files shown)


felangel:
Hi @MehadeHassan 👋 
Can you explain why you prefer this change? Both syntaxes are valid but the current is more concise and relies on type inference rather that explicitly having to specify the generic types.

MehadeHassan:
> Hi @MehadeHassan 
> Can you explain why you prefer this change? Both syntaxes are valid but the current is more concise and relies on type inference rather than explicitly having to specify the generic types.

It is true that the current one is concise. 

but, 
 1/ If we take a look into flutter_bloc's other `context` methods snippets, both `read` and `watch` use generic types. 
so, it will look more similar, if we use the same pattern (generic types). 

```
 // context.select<AuthenticationBloc, AuthenticationState>((bloc) => bloc.state),
 // context.read<AuthenticationBloc>().state,
 // context.watch<AuthenticationBloc>().state,
```
2/ I think the new snippet is more readable than the old one.

3/ Compatible with [avoid_types_on_closure_parameters](https://dart-lang.github.io/linter/lints/avoid_types_on_closure_parameters.html) lint rule.

PR #2789, opened 2021-09-24
Remove unneeded imports

kevmoo:
New hint!

### DIFF
--- a/examples/flutter_bloc_with_stream/lib/main.dart
+++ b/examples/flutter_bloc_with_stream/lib/main.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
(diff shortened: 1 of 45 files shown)


felangel (dismissed):
Awesome, thanks @kevmoo! 🎉 
Why weren't these being reported by the analyzer and what did you run to detect them `dart fix --apply`?

kevmoo:
@felangel – this is a BRAND new lint that's only on dev channel.

I don't think there's a quick fix for it – yet!

So you'll hit this when your tools update to latest. But it doesn't hurt to get a start now 😄

felangel:
> @felangel – this is a BRAND new lint that's only on dev channel.
> 
> I don't think there's a quick fix for it – yet!
> 
> So you'll hit this when your tools update to latest. But it doesn't hurt to get a start now 😄

Nice, thanks so much! 💙

felangel (approved):
🚀

PR #2815, opened 2021-09-30
refactor: clean unnecessary code on FilterButton

guillempuche:
## Status

**READY**

## Breaking Changes

NO

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/examples/flutter_firestore_todos/lib/widgets/filter_button.dart
+++ b/examples/flutter_firestore_todos/lib/widgets/filter_button.dart
@@ -10,11 +10,6 @@ class FilterButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final defaultStyle = Theme.of(context).textTheme.bodyText2;
-    final activeStyle = Theme.of(context)
-        .textTheme
-        .bodyText2
-        ?.copyWith(color: Theme.of(context).colorScheme.secondary);
     return BlocBuilder<FilteredTodosBloc, FilteredTodosState>(
         builder: (context, state) {
       final button = _Button(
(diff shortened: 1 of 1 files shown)


marcossevilla:
hey @guillempuche, it's okay to have parametrized the TextStyles for the private `_Button` widget so if we want to create another button using the private one and customizing it with other TextStyles. What do you think?

guillempuche:
True Marcos.

Sometimes clean is better than coding for a possible future. In [Unix philosophy](http://www.faqs.org/docs/artu/ch01s06.html) there's this:
> Rule of Extensibility: Design for the future, because it will be here sooner than you think. 

But it also says:
> Rule of Simplicity: Design for simplicity; add complexity only where you must. 

I think cleaning it is better for readability of the example. We can *refactor* on the future, instead of predicting.

It's up to you 🙂.

marcossevilla:
> True Marcos.
> 
> Sometimes clean is better than coding for a possible future. In [Unix philosophy](http://www.faqs.org/docs/artu/ch01s06.html) there's this:
> 
> > Rule of Extensibility: Design for the future, because it will be here sooner than you think.
> 
> But it also says:
> 
> > Rule of Simplicity: Design for simplicity; add complexity only where you must.
> 
> I think cleaning it is better for readability of the example. We can _refactor_ on the future, instead of predicting.
> 
> It's up to you 🙂.

I agree with refactoring in the future, but as this already was designed in a way you can create other buttons specifying other TextStyles, it would lead to a rollback to the current implementation if we want to use it later.

So if you agree, I'll close this PR as it's a minor change and the current implementation is simple enough IMO.

What do you think?

guillempuche:
If you want to reuse the button, why making it private? It indicates to me that its purpose is serving the Filter button.

Also, the _Button is very customized for the popup menu of Todos. If you want to be reused in the future, the `PopupMenuItem`, the type `VisibilityFilter` or the tooltip have to receive data from the parent widget, in my understanding. You adapt all the parts (or the important) for the reusability or nothing.

But this is a minor implementation. I close it. Thanks for the dialogue, Marcos 🙂.

PR #2896, opened 2021-10-22
feat(bloc)!: expose Emitter in Cubit

felangel:
## Status

**IN DEVELOPMENT**

## Breaking Changes

YES

## Description

- feat(bloc)!: expose Emitter in Cubit
  - support `emit.forEach` and `emit.onEach` in `Cubit`
- refactor(bloc)!: decouple `BlocBase` implementation from interface (closes ##2453)

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/docs/_snippets/testing/counter_bloc_test_initial_state.dart.md
+++ b/docs/_snippets/testing/counter_bloc_test_initial_state.dart.md
@@ -1,6 +1,6 @@
 ```dart
 group('CounterBloc', () {
-    CounterBloc counterBloc;
+    late CounterBloc counterBloc;
 
     setUp(() {
         counterBloc = CounterBloc();
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -42,4 +42,4 @@ We should be able to run the tests and see that all are passing.
 
 That's all there is to it, testing should be a breeze and we should feel confident when making changes and refactoring our code.
 
-You can refer to the [Todos App](https://github.com/brianegan/flutter_architecture_samples/tree/master/bloc_library) for an example of a fully tested application.
+You can refer to the [Weather App](https://github.com/felangel/bloc/tree/master/examples/flutter_weather) for an example of a fully tested application.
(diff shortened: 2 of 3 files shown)


marctan:
Hello, sorry for bumping this thread., but if I want to listen to a stream in Cubit class is below correct?

```
class ChatCubit extends Cubit<ChatState> {
  ChatCubit({required this.repository}) : super(const ChatState());
  final ChatRepository repository;
 StreamSubscription? subscription;

  @override
  Future<void> close() {
    subscription?.cancel();
    return super.close();
  }

void getMessages(String recipient) async {
    emit(state.copyWith(status: ChatStatus.messagesLoading));
    await subscription?.cancel();

    subscription = repository.getMessageStream(recipient).listen((event) {
      final List<ChatMessage> messageList = event.docs
          .map(
            (e) => ChatMessage(
              text: e['message'],
              user: ChatUser(
                uid: e['sender'],
                name: e['sender'],
              ),
              createdAt: DateTime.fromMillisecondsSinceEpoch(
                e['time'],
              ),
            ),
          )
          .toList();
      emit(state.copWith(status: ChatStatus.messagesLoaded, messages: messageList));
    });
  }
}
```

Thanks!
@felangel

Gene-Dana:
Feel free to put the subscription in the constructor body too!

zamaniafshar:
> * `emit.forEach` and `emit.onEach` while functional in Cubit will not be cancelable which imo reduces the usefulness significantly
> * The new API opens the door to developers being able to cache a reference of an `Emitter` since `emit` would be a getter
> * `isDone` only makes sense in the context of `Bloc` but is part of the base `Emitter` interface so we'd either need to have a specific `BlocEmitter` and `CubitEmitter` or `Cubit` users would have access to a useless property.
> 


I think the main reason this hasn’t been done yet is due to implementation challenges, not because adding these features to Cubit is a bad idea.
I really hope this feature is added to Bloc soon, as it’s a big pain to have to use Bloc in many situations just because of a few streams—which feels unreasonable. In the meantime, I've created a solution for myself.
https://github.com/felangel/bloc/issues/3555#issuecomment-2438463982

PR #2919, opened 2021-11-04
feat(bloc_test): prettier diffing

### LINKED ISSUE #1783: [Feature Request] Better diff between expected and actual in blocTest
Right now when my state differs from what I expect in bloc tests I get a big load of text on my screen showing what expected and actual states look like. The problem is, sometimes my state contains many items or the item itself is very big. It's **REALLY** hard to read it.

I can see two solutions:
- Showing a diff as GitHub does.
- Outputting only the parts of the object that really differed. So that when I have some object as a state, but only 2 out of 25 fields differ, I'd only care to see those two fields' values instead of the whole object.

Right now I just copy&paste both expected and actual states to some diff checker online to see the difference.

**Additional context**
Find difference with your eye :)

image

felangel:
## Status

**READY**

## Breaking Changes

NO

## Description

- feat(bloc_test): prettier diffing
  - closes #1783 

Screen Shot 2021-11-04 at 12 00 12 AM

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -247,28 +247,30 @@ String _stringDiff({required dynamic expected, required dynamic actual}) {
   buffer
     ..writeln('${"=" * 4} diff ${"=" * 40}')
     ..writeln('')
-    ..writeln('${_diffToString(differences: differences)}')
+    ..writeln(differences.toPrettyString())
     ..writeln('')
     ..writeln('${"=" * 4} end diff ${"=" * 36}');
   return buffer.toString();
 }
 
-String _diffToString({required List<Diff> differences}) {
-  String identical(String str) => '\u001b[90m$str\u001B[0m';
-  String deletion(String str) => '\u001b[31m[-$str-]\u001B[0m';
-  String insertion(String str) => '\u001b[32m{+$str+}\u001B[0m';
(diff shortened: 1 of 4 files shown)


packages/bloc_test/lib/src/bloc_test.dart
```diff
@@ -231,3 +240,37 @@ Alternatively, consider using Matchers in the expect of the blocTest rather than
...
+
+extension on List<Diff> {
+  String toPrettyString() {
+    String identical(String str) => '\u001b[90m$str\u001B[0m';
+    String deletion(String str) => '\u001b[31m[-$str-]\u001B[0m';
+    String insertion(String str) => '\u001b[32m{+$str+}\u001B[0m';
+
+    final buffer = StringBuffer();
+    for (final difference in this) {
+      if (difference.operation == DIFF_EQUAL) {
```
erickzanardo, line 264: nit: I believe operation is an enum? If so a switch case may be safer here?

erickzanardo (dismissed):

packages/bloc_test/lib/src/bloc_test.dart
felangel: Good call, refactored 👍

PR #2920, opened 2021-11-04
fix(bloc_listener) null pointer exception when no child has been given

devaronius:
## Status

**READY**

## Breaking Changes

NO

## Description

When you add a BlocListener to the Widget tree the child is optional. to not break the current BlocListener interface I've made a check on the overridden buildWithChild that if the child is null when returned a Sizedbox Shrink constructor is called to return a valid non-nullable widget. 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_listener.dart
+++ b/packages/flutter_bloc/lib/src/bloc_listener.dart
@@ -185,7 +185,7 @@ class _BlocListenerBaseState<B extends BlocBase<S>, S>
       // See https://github.com/felangel/bloc/issues/2127.
       context.select<B, bool>((bloc) => identical(_bloc, bloc));
     }
-    return child!;
+    return child ?? const SizedBox.shrink();
   }
 
   @override


felangel:
Hi @devaronius 👋 
Thanks for opening a PR! Can you please provide some more context regarding why you'd want to be able to use a `BlocListener` without a child? Do you have an example use-case in mind that you could share? Thanks!

devaronius:
> Hi @devaronius 👋 Thanks for opening a PR!
> 
> Can you please provide some more context regarding why you'd want to be able to use a `BlocListener` without a child? Do you have an example use-case in mind that you could share? Thanks!

hi @felangel,

For a client with an existing Flutter project, they had the following structure:

`MultiBlocProvider(
      providers: [
        BlocProvider<BlocA>(),
        BlocProvider<BlocB>(),
      ],
      child: MultiBlocListener(
        listeners: [
          BlocListener<BlocA>(),
          BlocListener<BlocB>(),
        ],
        child: BlocConsumer<BlocC, StateC>(
            builder: (context,state){
               Column(children: [
                   'The page widget'
               ValueListenableBuilder<bool>(
                        valueListenable: dicSetup.blocDReady,
                        builder: (context, isReady, _) {
                        isReady ? BlocListener<BlocD>() : SizedBox.Shrink(),
                        },
                    ),
               ],
             ),
            }
        ),
   );
   
   The page here is creating the "BlocD" on runtime and when it exist they want to listen to it for its states that are used for navigation purposes.
   
   I have to say, it was a first for me as well to use a blocListener this way, but in theory there is not really anything wrong with it.
   
   The problem only was the nullable to null cast that was done. for the MultiBlocListener it should be prefert not having a child. but in this context we found the 0.1% when we only want the listener without a child widget attached.
   
   If you have some more question, I love to provide you with more information.

felangel:
> > Hi @devaronius 👋 Thanks for opening a PR!
> > Can you please provide some more context regarding why you'd want to be able to use a `BlocListener` without a child? Do you have an example use-case in mind that you could share? Thanks!
> 
> hi @felangel,
> 
> For a client with an existing Flutter project, they had the following structure:
> 
> ```
> MultiBlocProvider(
>       providers: [
>         BlocProvider<BlocA>(),
>         BlocProvider<BlocB>(),
>       ],
>       child: MultiBlocListener(
>         listeners: [
>           BlocListener<BlocA>(),
>           BlocListener<BlocB>(),
>         ],
>         child: BlocConsumer<BlocC, StateC>(
>             builder: (context,state){
>                Column(children: [
>                    'The page widget'
>                ValueListenableBuilder<bool>(
>                         valueListenable: dicSetup.blocDReady,
>                         builder: (context, isReady, _) {
>                         return isReady ? BlocListener<BlocD>() : SizedBox.Shrink(),
>                         },
>                     ),
>                ],
>              ),
>             }
>         ),
>    );
> ```
> 
> The page here is creating the "BlocD" on runtime and when it exist they want to listen to it for its states that are used for navigation purposes.
> 
> I have to say, it was a first for me as well to use a blocListener this way, but in theory there is not really anything wrong with it.
> 
> The problem only was the nullable to non-null cast that was done. for the MultiBlocListener it should be prefert not having a child. but in this context we found the 0.1% when we only want the listener without a child widget attached inside the widget tree itself.
> 
> If you have some more question, I love to provide you with more information.

Thanks for providing more context! As you mentioned, the reason why `child` is nullable in `BlocListener` is because a `child` should not be specified when using `BlocListener` within a `MultiBlocListener`. For consistency with `provider`, I'd prefer to throw an `AssertionError` if `BlocListener` does not specify a child rather than defaulting to a `SizedBox`.

In the example you provided, I'd recommend lifting the `BlocListener<BlocD>` up higher rather than having it be a leaf node in the sub-tree. If we allow `BlocListener` to be instantiated without a `child` we'd also have to allow `MultiBlocListener` to be instantiated without a child and similarly, `BlocProvider` has an optional `child` because it can be used within a `MultiBlocProvider`.

Would you be alright if rather than defaulting to a `SizedBox` we explicitly assert that `child != null` to align with how `provider` handles this scenario? Thanks!

felangel:
Closing for now since this behavior was addressed in #2924. Feel free to comment with additional questions/comments and I'm happy to continue the conversation 👍

Issue #2925, opened 2021-11-06
fix: [VS Code Bloc extension] When wrapp with bloc snippet, the $ expression in the widget content will be removed.


Nomeleel:
**Description**
If there is a $ expression in the widget, after wrap with bloc snippet, the expression will be removed and become an empty string.

**Steps To Reproduce**

1. Widget content
``` dart
Center(
  child: Column(
    mainAxisAlignment: MainAxisAlignment.center,
    children: <Widget>[
      Text(
        'You have pushed the button this many times:',
      ),
      Text(
        '$_counter',  // Here
        style: Theme.of(context).textTheme.headline4,
      ),
    ],
  ),
)
```
2.  Refactor -> Wrap with BlocSelector
```dart
BlocSelector<SubjectBloc, SubjectState, SelectedState>(
  selector: (state) {
    return state;
  },
  builder: (context, state) {
    return Center(
      child: Column(
        mainAxisAlignment: MainAxisAlignment.center,
        children: <Widget>[
          Text(
            'You have pushed the button this many times:',
          ),
          Text(
            '',  // Here
            style: Theme.of(context).textTheme.headline4,
          ),
        ],
      ),
    );
  },
)
```
3.  See code near the 'Here' comment

'$_counter' -->  ''

**Expected Behavior**
```dart
BlocSelector<SubjectBloc, SubjectState, SelectedState>(
  selector: (state) {
    return state;
  },
  builder: (context, state) {
    return Center(
      child: Column(
        mainAxisAlignment: MainAxisAlignment.center,
        children: <Widget>[
          Text(
            'You have pushed the button this many times:',
          ),
          Text(
            '$_counter',  // Here
            style: Theme.of(context).textTheme.headline4,
          ),
        ],
      ),
    );
  },
)
```

**Screenshots**

![Screenshots](https://raw.githubusercontent.com/Nomeleel/Assets/master/issue/bloc-vscode_wrap_issue.gif)

PR #2926, opened 2021-11-06
Fixed  vscode wrap issue.

Nomeleel:
## Status

FINISH

## Breaking Changes

 NO

## Description

When wrapp with bloc snippet, the $ expression in the widget content will be removed.

***Root case:*** VS Code insertSnippet will execute the $ expression in the text.

***Fix:*** Escape the $ to prevent it from being executed incorrectly.

## Type of Change
- [ x ] 🛠️ Bug fix (non-breaking change which fixes an issue)
#2925


felangel (approved):
LGTM ✅ 
Thanks so much for the contribution! 💯 💙

PR #2932, opened 2021-11-09
feat(bloc)!: BlocOverrides API

felangel:
## Status

**READY**

## Breaking Changes

YES

## Description

The current API used to override the default `BlocObserver` and `EventTransformer` is:

```dart
void main() {
  Bloc.observer = CustomBlocObserver();
  Bloc.transformer = customEventTransformer();
  
  // ...
}
```

This approach relies on a global singleton for both the `BlocObserver` and `EventTransformer`. As a result, it's not possible to:

- Have multiple `BlocObserver` or `EventTransformer` implementations scoped to different parts of the application
- Have `BlocObserver` or `EventTransformer` overrides be scoped to a package
  - Suppose a package that depends on `package:bloc` registers its own `BlocObserver`, any consumer of the package will either have to overwrite the package's `BlocObserver` or report to the package's `BlocObserver`.

It's also more difficult to test because of the shared, global state across tests.

This pull request introduces a `BlocOverrides` class which allows developers to override `BlocObserver` and/or `EventTransformer` for a specific `Zone` rather than relying on a global, mutable singleton.

```dart
void main() {
  BlocOverrides.runZoned(
    () {
      // ...
    },
    blocObserver: CustomBlocObserver(),
    eventTransformer: customEventTransformer(),
  );
}
```

Bloc instances will use the `BlocObserver` and/or `EventTransformer` for the current `Zone` via `BlocOverrides.current`. If there are no `BlocOverrides` for the zone, they will use the existing internal defaults (base `BlocObserver` and concurrent `EventTransformer` -- no change in terms of behavior/functionality).

As previously mentioned, this would allow each zone to function independently with its own `BlocOverrides` and `BlocOverrides` are immutable meaning the `BlocOverrides` specified when the zone is created cannot be modified.

```dart
BlocOverrides.runZoned(
  () {
    // BlocObserverA and eventTransformerA
    final overrides = BlocOverrides.current;
      
    // Blocs in this zone report to BlocObserverA
    // and use eventTransformerA as the default transformer.
    // ...

    // Later...
    BlocOverrides.runZoned(
      () {
        // BlocObserverB and eventTransformerB
        final overrides = BlocOverrides.current;
    
        // Blocs in this zone report to BlocObserverB
        // and use eventTransformerB as the default transformer.
        // ...
      },
      blocObserver: BlocObserverB(),
      eventTransformer: eventTransformerB(),
    );
  },
  blocObserver: BlocObserverA(),
  eventTransformer: eventTransformerA(),
);
```

This pull request removes the static, mutable `Bloc.observer` and `Bloc.transformer` APIs which is a breaking change. `BlocObserver` and `EventTransformer` overrides will need to be specified via `BlocOverrides.runZoned` instead as described above.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/bloc/test/cubit_test.dart
```diff
@@ -262,14 +275,15 @@ void main() {
...
         observer = MockBlocObserver();
-        Bloc.observer = observer;
       });
 
       test('triggers onClose on observer', () async {
-        final cubit = CounterCubit();
-        await cubit.close();
-        // ignore: invalid_use_of_protected_member
-        verify(() => observer.onClose(cubit)).called(1);
+        await BlocOverrides.runZoned(() async {
```
lhmzhou, line 281: Should we have some error handling for this async call?

felangel: It’s not necessary since this is a year and an unhandled exception will result in a test failure

lhmzhou (approved):
LGTM :shipit:

PR #2959, opened 2021-11-17
feat(flutter_bloc): Adds nullable return widget for BlocBuilder

Rukkaitto:
## Status

**READY**

## Breaking Changes

NO

## Description

When using BlocBuilder, there are a lot of times where you have to check your state's type and return an empty Container when none of the types are matched. This pull request makes the return Widget of the builder function nullable. If no widget is returned, the BlocBuilder will return an empty Container.

## Type of Change

- [x ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_builder.dart
+++ b/packages/flutter_bloc/lib/src/bloc_builder.dart
@@ -4,7 +4,7 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 /// Signature for the `builder` function which takes the `BuildContext` and
 /// [state] and is responsible for returning a widget which is to be rendered.
 /// This is analogous to the `builder` function in [StreamBuilder].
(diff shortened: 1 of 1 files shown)


felangel:
Hi @Rukkaitto 👋 
Thanks for opening a pull request. I would prefer not to make this change.

Allowing `BlocBuilder` to return `null` diverges from existing widgets like: `StreamBuilder`, `AnimatedBuilder`, `ValueListenableBuilder` etc. In addition, this change leaves the door open to treat `BlocBuilder` as a place for "side effects" because you don't have to return a widget which muddies the waters. I feel strongly that the `BlocBuilder` API should be strict and require that a widget is always returned explicitly. If you have a specific situation in mind that you feel is problematic feel free to provide a minimal reproduction sample, open an issue, and I'm happy to discuss this further.

Thanks!

PR #3012, opened 2021-11-28
refactor(bloc): add core interfaces

### LINKED ISSUE #2994: refactor: аdding interfaces
Why does the entire **architectural** library still lack a interfaces?
Whats about `Dependency inversion principle` and `interface segregation principle`?

Why `BlocBuilder` depends on `BlocBase` and not on `IStateObservable` something like this:
```dart
/// {@template i_state_observable}
/// An interface to observe states stream.
/// Current [state] available as property.
/// State subject to implement the Observer (Dependents) pattern
/// {@endtemplate}
abstract class IStateObservable<T extends Object?> {
  /// The current [state].
  T get state;

  /// The state stream.
  StateStream<T> get stream;
}
```
`BlocBuilder`, `BlocListener`, `BlocProvider`, `BlocConsumer` really needs `addError`, `emit`, `onChange`, `onError` and others methods and callbacks?

Also you need something like this:
```dart
/// {@template i_bloc_subject}
/// Interface to observe states stream.
/// Current [state] available as property.
/// [Bloc] Subject to implement the Publisher/Subscriber pattern
/// {@endtemplate}
abstract class IBlocSubject<State extends Object?>
    implements IStateObservable<State> {
  /// Updates the [state] to the provided [state].
  /// [setState] does nothing if the instance has been closed.
  @protected
  @mustCallSuper
  void setState(State state);

  /// Called whenever a [change] occurs with the given [change].
  /// A [change] occurs when a new `state` is emitted.
  /// [onChange] is called before the `state` of the `bloc` is updated.
  /// [onChange] is a great spot to add logging/analytics for a specific `bloc`.
  ///
  /// **Note: `super.onChange` should always be called first.**
  /// ```dart
  /// @override
  /// void onChange(Change change) {
  ///   // Always call super.onChange with the current change
  ///   super.onChange(change);
  ///
  ///   // Custom onChange logic goes here
  /// }
  /// ```
  ///
  /// See also:
  ///
  /// * [IBlocObserver] for observing [Bloc] behavior globally.
  @protected
  @mustCallSuper
  @visibleForOverriding
  void onChange(Change<State> change);
}

/// {@template i_bloc_sink}
/// A [Bloc] Sink interface that accepts events both synchronously
/// and asynchronously.
///
/// The [IBlocSink] methods can't be used while the [addStream] is called.
/// As soon as the [addStream]'s [Future] completes with a value, the
/// [EventSink] methods can be used again.
///
/// If [addStream] is called after any of the [EventSink] methods, it'll
/// be delayed until the underlying system has consumed the data added by the
/// [EventSink] methods.
///
/// When [EventSink] methods are used, the [done] [Future] can be used to
/// catch any errors.
///
/// When [close] is called, it will return the [done] [Future].
/// {@endtemplate}
abstract class IBlocSink<Event extends Object?> implements StreamSink<Event> {
  /// Reports an [error] which triggers [onError] with an optional [StackTrace].
  /// If [close] has already been called, any subsequent calls to [addError]
  /// will be ignored and will not result in any subsequent state changes.
  @override
  @protected
  @mustCallSuper

  /// Notifies the [Bloc] of a new [event]
  /// If [close] has already been called, any subsequent calls to [add] will
  /// be ignored and will not result in any subsequent state changes.
  @override
  void add(Event event);

  /// Called whenever an [event] is [add]ed to the [Bloc].
  /// A great spot to add logging/analytics at the individual [Bloc] level.
  ///
  /// **Note: `super.onEvent` should always be called first.**
  /// ```dart
  /// @override
  /// void onEvent(Event event) {
  ///   // Always call super.onEvent with the current event
  ///   super.onEvent(event);
  ///
  ///   // Custom onEvent logic goes here
  /// }
  /// ```
  ///
  /// See also:
  ///
  /// * [IBlocObserver.onEvent] for observing events globally.
  ///
  @protected
  @mustCallSuper
  @visibleForOverriding
  void onEvent(Event event);

  /// Called whenever an [error] occurs and notifies [IBlocObserver.onError].
  ///
  /// In debug mode, [onError] throws a [BlocUnhandledErrorException] for
  /// improved visibility.
  ///
  /// In release mode, [onError] does not throw and will instead only report
  /// the error to [IBlocObserver.onError].
  ///
  /// **Note: `super.onError` should always be called last.**
  /// ```dart
  /// @override
  /// void onError(Object error, StackTrace stackTrace) {
  ///   // Custom onError logic goes here
  ///
  ///   // Always call super.onError with the current error and stackTrace
  ///   super.onError(error, stackTrace);
  /// }
  /// ```
  @protected
  @mustCallSuper
  @visibleForOverriding
  void onError(Object error, StackTrace stackTrace);

  /// Closes the `event` and `state` `Streams`.
  /// This method should be called when a [Bloc] is no longer needed.
  /// Once [close] is called, `events` that are [add]ed will not be
  /// processed.
  /// In addition, if [close] is called while `events` are still being
  /// processed, the [Bloc] will finish processing the pending `events`.
  @override
  @mustCallSuper
  @visibleForOverriding
  Future<void> close();

  /// Pass all [eventStream] events to the [Bloc]
  /// If [close] has already been called, any subsequent calls to [addStream]
  /// will be ignored and will not result in any subsequent state changes.
  @override
  Future<void> addStream(Stream<Event> eventStream);

  /// Return a future which is completed when the [IBlocSink] is finished.
  ///
  /// If the `IBlocSink` fails with an error,
  /// perhaps in response to adding events using [add], [addStream], [addError]
  /// or [close], the [done] future will complete with that error.
  ///
  /// Otherwise, the returned future will complete when either:
  ///
  /// * all events have been processed and the sink has been closed, or
  /// * the sink has otherwise been stopped from handling more events
  ///   (for example by canceling a stream subscription).
  @override
  Future<void> get done;
}

/// {@template i_bloc}
/// An interface to implement the Publisher/Subscriber pattern for [Bloc].
/// Takes a `Stream` of `Events` as input
/// and transforms them into a `Stream` of `States` as output.
/// {@endtemplate}
abstract class IBloc<Event extends Object?, State extends Object?>
    implements IBlocSink<Event>, IBlocSubject<State> {
  /// Transforms the [events] stream along with a [transitionFn] function into
  /// a `Stream<Transition>`.
  /// Events that should be processed by [mapEventToState] need to be passed to
  /// [transitionFn].
  /// By default `asyncExpand` is used to ensure all [events] are processed in
  /// the order in which they are received.
  /// You can override [transformEvents] for advanced usage in order to
  /// manipulate the frequency and specificity with which [mapEventToState] is
  /// called as well as which [events] are processed.
  ///
  /// For example, if you only want [mapEventToState] to be called on the most
  /// recent [Event] you can use `switchMap` instead of `asyncExpand`.
  ///
  /// ```dart
  /// @override
  /// Stream<Transition<Event, State>> transformEvents(events, transitionFn) {
  ///   return events.switchMap(transitionFn);
  /// }
  /// ```
  ///
  /// Alternatively, if you only want [mapEventToState] to be called for
  /// distinct [events]:
  ///
  /// ```dart
  /// @override
  /// Stream<Transition<Event, State>> transformEvents(events, transitionFn) {
  ///   return super.transformEvents(
  ///     events.distinct(),
  ///     transitionFn,
  ///   );
  /// }
  /// ```
  @visibleForOverriding
  Stream<Transition<Event, State>> transformEvents(
    Stream<Event> events,
    TransitionFunction<Event, State> transitionFn,
  );

  /// Must be implemented when a class extends [Bloc].
  /// [mapEventToState] is called whenever an [event] is [add]ed
  /// and is responsible for converting that [event] into a new [state].
  /// [mapEventToState] can `yield` zero, one, or multiple states for an event.
  @protected
  @visibleForOverriding
  Stream<State> mapEventToState(Event event);

  /// Called whenever a [transition] occurs with the given [transition].
  /// A [transition] occurs when a new `event` is [add]ed and [mapEventToState]
  /// executed.
  /// [onTransition] is called before a [Bloc]'s [state] has been updated.
  /// A great spot to add logging/analytics at the individual [Bloc] level.
  ///
  /// **Note: `super.onTransition` should always be called first.**
  /// ```dart
  /// @override
  /// void onTransition(Transition<Event, State> transition) {
  ///   // Always call super.onTransition with the current transition
  ///   super.onTransition(transition);
  ///
  ///   // Custom onTransition logic goes here
  /// }
  /// ```
  ///
  /// See also:
  ///
  /// * [IBlocObserver.onTransition] for observing transitions globally.
  ///
  @protected
  @mustCallSuper
  @visibleForOverriding
  void onTransition(Transition<Event, State> transition);

  /// Transforms the `Stream<Transition>` into a new `Stream<Transition>`.
  /// By default [transformTransitions] returns
  /// the incoming `Stream<Transition>`.
  /// You can override [transformTransitions] for advanced usage in order to
  /// manipulate the frequency and specificity at which `transitions`
  /// (state changes) occur.
  ///
  /// For example, if you want to debounce outgoing state changes:
  ///
  /// ```dart
  /// @override
  /// Stream<Transition<Event, State>> transformTransitions(
  ///   Stream<Transition<Event, State>> transitions,
  /// ) {
  ///   return transitions.debounceTime(Duration(seconds: 1));
  /// }
  /// ```
  @protected
  @visibleForOverriding
  Stream<Transition<Event, State>> transformTransitions(
    Stream<Transition<Event, State>> transitions,
  );
}
```

and much more, like `ISink`, `IDisposal`, `ITransition`, `IChange`, `IEventConsumer`, `IStateEmiter`.

felangel:
## Status

**READY**

## Breaking Changes

NO

## Description

- refactor(bloc): add core interfaces (closes #2994)

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -1,5 +1,15 @@
 part of bloc;
 
+/// An [ErrorSink] that supports adding events.
+///
+/// Multiple events can be reported to the sink via `add`.
+abstract class EventSink<Event extends Object?> implements ErrorSink {
+  /// Adds an [event] to the sink.
+  ///
+  /// Must not be called on a closed sink.
+  void add(Event event);
+}
+
 /// An event handler is responsible for reacting to an incoming [Event]
 /// and can emit zero or more states via the [Emitter].
 typedef EventHandler<Event, State> = FutureOr<void> Function(
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -29,7 +29,8 @@ class _Handler {
 /// Takes a `Stream` of `Events` as input
 /// and transforms them into a `Stream` of `States` as output.
 /// {@endtemplate}
-abstract class Bloc<Event, State> extends BlocBase<State> {
+abstract class Bloc<Event, State> extends BlocBase<State>
+    implements EventSink<Event> {
   /// {@macro bloc}
   Bloc(State initialState) : super(initialState);
 
@@ -48,6 +49,7 @@ abstract class Bloc<Event, State> extends BlocBase<State> {
   ///
   /// * A [StateError] will be thrown if the bloc is closed and the
   /// [event] will not be processed.
+  @override
   void add(Event event) {
     assert(() {
       final handlerExists = _handlers.any((handler) => handler.isType(event));
@@ -240,8 +242,8 @@ abstract class Bloc<Event, State> extends BlocBase<State> {
   /// processed.
   /// In addition, if [close] is called while `events` are still being
   /// processed, the [Bloc] will finish processing the pending `events`.
-  @override
   @mustCallSuper
+  @override
   Future<void> close() async {
     await _eventController.close();
     for (final emitter in _emitters) emitter.cancel();
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -10,7 +10,7 @@ part 'emitter.dart';
 /// An [ErrorSink] that supports adding events.
 ///
 /// Multiple events can be reported to the sink via `add`.
-abstract class EventSink<Event extends Object?> implements ErrorSink {
+abstract class BlocEventSink<Event extends Object?> implements ErrorSink {
   /// Adds an [event] to the sink.
   ///
   /// Must not be called on a closed sink.
@@ -47,7 +47,7 @@ class _Handler {
 /// and transforms them into a `Stream` of `States` as output.
 /// {@endtemplate}
 abstract class Bloc<Event, State> extends BlocBase<State>
-    implements EventSink<Event> {
+    implements BlocEventSink<Event> {
   /// {@macro bloc}
   Bloc(State initialState) : super(initialState);
 
--- a/packages/bloc/lib/src/bloc_base.dart
+++ b/packages/bloc/lib/src/bloc_base.dart
@@ -1,19 +1,19 @@
 part of bloc;
 
 /// An object that provides access to a stream of states over time.
-abstract class Streamable<State> {
+abstract class Streamable<State extends Object?> {
   /// The current [stream] of states.
   Stream<State> get stream;
 }
 
 /// A [Streamable] that provides synchronous access to the current [state].
-abstract class StateStreamable<State> extends Streamable<State> {
+abstract class StateStreamable<State> implements Streamable<State> {
   /// The current [state].
   State get state;
 }
 
 /// An object which can emit new states.
-abstract class Emittable<State> {
+abstract class Emittable<State extends Object?> {
   /// Emits a new [state].
   void emit(State state);
 }
@@ -22,13 +22,13 @@ abstract class Emittable<State> {
 abstract class Closable {
   /// Closes the current instance.
   /// The returned future completes when the instance has been closed.
-  Future<void> close();
+  FutureOr<void> close();
 }
 
 /// A generic destination for errors.
 ///
 /// Multiple errors can be reported to the sink via `addError`.
-abstract class ErrorSink extends Closable {
+abstract class ErrorSink implements Closable {
   /// Adds an [error] to the sink with an optional [stackTrace].
   ///
   /// Must not be called on a closed sink.
--- a/packages/bloc/lib/src/bloc_base.dart
+++ b/packages/bloc/lib/src/bloc_base.dart
@@ -1,10 +1,46 @@
 part of bloc;
 
+/// An object that provides access to a stream of states over time.
+abstract class Streamable<State> {
+  /// The current [stream] of states.
+  Stream<State> get stream;
+}
+
+/// A [Streamable] that provides synchronous access to the current [state].
+abstract class StateStreamable<State> extends Streamable<State> {
+  /// The current [state].
+  State get state;
+}
+
+/// An object which can emit new states.
+abstract class Emittable<State> {
+  /// Emits a new [state].
+  void emit(State state);
+}
+
+/// An object which must be closed when no longer in use.
+abstract class Closable {
+  /// Closes the current instance.
+  /// The returned future completes when the instance has been closed.
+  Future<void> close();
+}
+
+/// A generic destination for errors.
+///
+/// Multiple errors can be reported to the sink via `addError`.
+abstract class ErrorSink extends Closable {
+  /// Adds an [error] to the sink with an optional [stackTrace].
+  ///
+  /// Must not be called on a closed sink.
+  void addError(Object error, [StackTrace? stackTrace]);
+}
+
 /// {@template bloc_base}
 /// An interface for the core functionality implemented by
 /// both [Bloc] and [Cubit].
 /// {@endtemplate}
-abstract class BlocBase<State> {
+abstract class BlocBase<State>
+    implements StateStreamable<State>, Emittable<State>, ErrorSink {
   /// {@macro bloc_base}
   BlocBase(this._state) {
     // ignore: invalid_use_of_protected_member
@@ -22,10 +58,10 @@ abstract class BlocBase<State> {
 
   bool _emitted = false;
 
-  /// The current [state].
+  @override
   State get state => _state;
 
-  /// The current state stream.
+  @override
   Stream<State> get stream => _stateController.stream;
 
   /// Whether the bloc is closed.
@@ -45,6 +81,7 @@ abstract class BlocBase<State> {
   /// * Throws a [StateError] if the bloc is closed.
   @protected
   @visibleForTesting
+  @override
   void emit(State state) {
     try {
       if (isClosed) {
@@ -89,6 +126,7 @@ abstract class BlocBase<State> {
 
   /// Reports an [error] which triggers [onError] with an optional [StackTrace].
   @mustCallSuper
+  @override
   void addError(Object error, [StackTrace? stackTrace]) {
     onError(error, stackTrace ?? StackTrace.current);
   }
@@ -117,6 +155,7 @@ abstract class BlocBase<State> {
   /// This method should be called when the instance is no longer needed.
   /// Once [close] is called, the instance can no longer be used.
   @mustCallSuper
+  @override
   Future<void> close() async {
     // ignore: invalid_use_of_protected_member
     _blocObserver?.onClose(this);
--- a/packages/bloc/lib/src/bloc_base.dart
+++ b/packages/bloc/lib/src/bloc_base.dart
@@ -12,17 +12,21 @@ abstract class StateStreamable<State> implements Streamable<State> {
   State get state;
 }
 
-/// An object which can emit new states.
-abstract class Emittable<State extends Object?> {
-  /// Emits a new [state].
-  void emit(State state);
-}
+/// A [StateStreamable] that must be closed when no longer in use.
+abstract class StateStreamableSource<State>
+    implements StateStreamable<State>, Closable {}
 
-/// An object which must be closed when no longer in use.
+/// An object that must be closed when no longer in use.
 abstract class Closable {
   /// Closes the current instance.
   /// The returned future completes when the instance has been closed.
   FutureOr<void> close();
+}
+
+/// An object that can emit new states.
+abstract class Emittable<State extends Object?> {
+  /// Emits a new [state].
+  void emit(State state);
 }
 
 /// A generic destination for errors.
@@ -40,7 +44,7 @@ abstract class ErrorSink implements Closable {
 /// both [Bloc] and [Cubit].
 /// {@endtemplate}
 abstract class BlocBase<State>
-    implements StateStreamable<State>, Emittable<State>, ErrorSink {
+    implements StateStreamableSource<State>, Emittable<State>, ErrorSink {
   /// {@macro bloc_base}
   BlocBase(this._state) {
     // ignore: invalid_use_of_protected_member
--- a/packages/flutter_bloc/lib/src/bloc_builder.dart
+++ b/packages/flutter_bloc/lib/src/bloc_builder.dart
@@ -69,7 +69,8 @@ typedef BlocBuilderCondition<S> = bool Function(S previous, S current);
 /// )
 /// ```
 /// {@endtemplate}
-class BlocBuilder<B extends BlocBase<S>, S> extends BlocBuilderBase<B, S> {
+class BlocBuilder<B extends StateStreamable<S>, S>
+    extends BlocBuilderBase<B, S> {
   /// {@macro bloc_builder}
   /// {@macro bloc_builder_build_when}
   const BlocBuilder({
@@ -97,7 +98,7 @@ class BlocBuilder<B extends BlocBase<S>, S> extends BlocBuilderBase<B, S> {
 /// so far. The type of the state and how it is updated with each interaction
 /// is defined by sub-classes.
 /// {@endtemplate}
-abstract class BlocBuilderBase<B extends BlocBase<S>, S>
+abstract class BlocBuilderBase<B extends StateStreamable<S>, S>
     extends StatefulWidget {
   /// {@macro bloc_builder_base}
   const BlocBuilderBase({Key? key, this.bloc, this.buildWhen})
@@ -118,7 +119,7 @@ abstract class BlocBuilderBase<B extends BlocBase<S>, S>
   State<BlocBuilderBase<B, S>> createState() => _BlocBuilderBaseState<B, S>();
 }
 
-class _BlocBuilderBaseState<B extends BlocBase<S>, S>
+class _BlocBuilderBaseState<B extends StateStreamable<S>, S>
     extends State<BlocBuilderBase<B, S>> {
   late B _bloc;
   late S _state;
--- a/packages/flutter_bloc/lib/src/bloc_provider.dart
+++ b/packages/flutter_bloc/lib/src/bloc_provider.dart
@@ -34,7 +34,7 @@ mixin BlocProviderSingleChildWidget on SingleChildWidget {}
 /// ```
 ///
 /// {@endtemplate}
-class BlocProvider<T extends StateStreamable<Object?>>
+class BlocProvider<T extends StateStreamableSource<Object?>>
     extends SingleChildStatelessWidget with BlocProviderSingleChildWidget {
   /// {@macro bloc_provider}
   const BlocProvider({
@@ -92,7 +92,7 @@ class BlocProvider<T extends StateStreamable<Object?>>
   /// ```dart
   /// BlocProvider.of<BlocA>(context);
   /// ```
-  static T of<T extends StateStreamable<Object?>>(
+  static T of<T extends StateStreamableSource<Object?>>(
     BuildContext context, {
     bool listen = false,
   }) {
@@ -129,11 +129,7 @@ class BlocProvider<T extends StateStreamable<Object?>>
           )
         : InheritedProvider<T>(
             create: _create,
-            dispose: (_, bloc) {
-              if (bloc is Closable) {
-                (bloc as Closable).close();
-              }
-            },
+            dispose: (_, bloc) => bloc.close(),
             startListening: _startListening,
             child: child,
             lazy: lazy,
--- a/packages/flutter_bloc/lib/src/bloc_provider.dart
+++ b/packages/flutter_bloc/lib/src/bloc_provider.dart
@@ -34,7 +34,7 @@ mixin BlocProviderSingleChildWidget on SingleChildWidget {}
 /// ```
 ///
 /// {@endtemplate}
-class BlocProvider<T extends BlocBase<Object?>>
+class BlocProvider<T extends StateStreamable<Object?>>
     extends SingleChildStatelessWidget with BlocProviderSingleChildWidget {
   /// {@macro bloc_provider}
   const BlocProvider({
@@ -92,7 +92,7 @@ class BlocProvider<T extends BlocBase<Object?>>
   /// ```dart
   /// BlocProvider.of<BlocA>(context);
   /// ```
-  static T of<T extends BlocBase<Object?>>(
+  static T of<T extends StateStreamable<Object?>>(
     BuildContext context, {
     bool listen = false,
   }) {
@@ -129,7 +129,11 @@ class BlocProvider<T extends BlocBase<Object?>>
           )
         : InheritedProvider<T>(
             create: _create,
-            dispose: (_, bloc) => bloc.close(),
+            dispose: (_, bloc) {
+              if (bloc is Closable) {
+                (bloc as Closable).close();
+              }
+            },
             startListening: _startListening,
             child: child,
             lazy: lazy,
@@ -137,8 +141,8 @@ class BlocProvider<T extends BlocBase<Object?>>
   }
 
   static VoidCallback _startListening(
-    InheritedContext<BlocBase?> e,
-    BlocBase value,
+    InheritedContext<StateStreamable?> e,
+    StateStreamable value,
   ) {
     final subscription = value.stream.listen(
       (dynamic _) => e.markNeedsNotifyDependents(),
--- a/packages/flutter_bloc/lib/src/bloc_consumer.dart
+++ b/packages/flutter_bloc/lib/src/bloc_consumer.dart
@@ -58,7 +58,7 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 /// )
 /// ```
 /// {@endtemplate}
-class BlocConsumer<B extends BlocBase<S>, S> extends StatefulWidget {
+class BlocConsumer<B extends StateStreamable<S>, S> extends StatefulWidget {
   /// {@macro bloc_consumer}
   const BlocConsumer({
     Key? key,
@@ -98,7 +98,7 @@ class BlocConsumer<B extends BlocBase<S>, S> extends StatefulWidget {
   State<BlocConsumer<B, S>> createState() => _BlocConsumerState<B, S>();
 }
 
-class _BlocConsumerState<B extends BlocBase<S>, S>
+class _BlocConsumerState<B extends StateStreamable<S>, S>
     extends State<BlocConsumer<B, S>> {
   late B _bloc;
 
--- a/packages/flutter_bloc/lib/src/bloc_listener.dart
+++ b/packages/flutter_bloc/lib/src/bloc_listener.dart
@@ -77,8 +77,8 @@ typedef BlocListenerCondition<S> = bool Function(S previous, S current);
 /// )
 /// ```
 /// {@endtemplate}
-class BlocListener<B extends BlocBase<S>, S> extends BlocListenerBase<B, S>
-    with BlocListenerSingleChildWidget {
+class BlocListener<B extends StateStreamable<S>, S>
+    extends BlocListenerBase<B, S> with BlocListenerSingleChildWidget {
   /// {@macro bloc_listener}
   /// {@macro bloc_listener_listen_when}
   const BlocListener({
@@ -103,7 +103,7 @@ class BlocListener<B extends BlocBase<S>, S> extends BlocListenerBase<B, S>
 /// The type of the state and what happens with each state change
 /// is defined by sub-classes.
 /// {@endtemplate}
-abstract class BlocListenerBase<B extends BlocBase<S>, S>
+abstract class BlocListenerBase<B extends StateStreamable<S>, S>
     extends SingleChildStatefulWidget {
   /// {@macro bloc_listener_base}
   const BlocListenerBase({
@@ -135,7 +135,7 @@ abstract class BlocListenerBase<B extends BlocBase<S>, S>
       _BlocListenerBaseState<B, S>();
 }
 
-class _BlocListenerBaseState<B extends BlocBase<S>, S>
+class _BlocListenerBaseState<B extends StateStreamable<S>, S>
     extends SingleChildState<BlocListenerBase<B, S>> {
   StreamSubscription<S>? _subscription;
   late B _bloc;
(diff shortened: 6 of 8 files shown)


packages/bloc/lib/src/bloc_base.dart
```diff
@@ -1,10 +1,46 @@
 part of bloc;
 
+/// An object that provides access to a stream of states over time.
+abstract class Streamable<State> {
+  /// The current [stream] of states.
+  Stream<State> get stream;
+}
+
+/// A [Streamable] that provides synchronous access to the current [state].
+abstract class StateStreamable<State> extends Streamable<State> {
```
PlugFox, line 10: Must be `implement Streamable<State>`

packages/bloc/lib/src/bloc_base.dart
```diff
@@ -1,10 +1,46 @@
...
+abstract class Closable {
+  /// Closes the current instance.
+  /// The returned future completes when the instance has been closed.
+  Future<void> close();
+}
+
+/// A generic destination for errors.
+///
+/// Multiple errors can be reported to the sink via `addError`.
+abstract class ErrorSink extends Closable {
```
PlugFox, line 31: Must be `implement Closable`

packages/bloc/lib/src/bloc_base.dart
```diff
@@ -1,10 +1,46 @@
 part of bloc;
 
+/// An object that provides access to a stream of states over time.
+abstract class Streamable<State> {
```
PlugFox, line 4: This is only style suggestion:
I like type generics like `<State extends Object>` if anything except `Null` and `Void`
or `<State extends Object?>` if litteraly anything

packages/bloc/lib/src/bloc_base.dart
```diff
@@ -1,10 +1,46 @@
...
+}
+
+/// A [Streamable] that provides synchronous access to the current [state].
+abstract class StateStreamable<State> extends Streamable<State> {
+  /// The current [state].
+  State get state;
+}
+
+/// An object which can emit new states.
+abstract class Emittable<State> {
```
PlugFox, line 16: This is only style suggestion:
I like type generics like `<State extends Object>` if anything except Null and Void
or `<State extends Object?>` if litteraly anything

packages/bloc/lib/src/bloc_base.dart
```diff
@@ -1,10 +1,46 @@
...
+  /// Must not be called on a closed sink.
+  void addError(Object error, [StackTrace? stackTrace]);
+}
+
 /// {@template bloc_base}
 /// An interface for the core functionality implemented by
 /// both [Bloc] and [Cubit].
 /// {@endtemplate}
-abstract class BlocBase<State> {
+abstract class BlocBase<State>
```
PlugFox, line 42: This is only style suggestion:
I like type generics like `<State extends Object>` if anything except Null and Void
or `<State extends Object?>` if litteraly anything

packages/bloc/lib/src/bloc_base.dart
```diff
@@ -1,10 +1,46 @@
...
+abstract class Emittable<State> {
+  /// Emits a new [state].
+  void emit(State state);
+}
+
+/// An object which must be closed when no longer in use.
+abstract class Closable {
+  /// Closes the current instance.
+  /// The returned future completes when the instance has been closed.
+  Future<void> close();
```
PlugFox, line 25: This is only style suggestion:
Maybe use `FutureOr<void> close()`

PlugFox:
I'm sorry for many reviews instead of one :'C
I'll do better next time

packages/bloc/lib/src/bloc_base.dart
felangel: That will technically be a breaking change at this point because currently you can create a Bloc with a dynamic state. I'm not opposed to making the change in the future though but I don't think it should be done at this point because it is technically breaking.

PlugFox: Missundestending)
You can set `BlocBase<State extends Object?>` without breaking changes.
I would do like this. But this is only style point.
If you don’t like it, don’t need to, of course

packages/bloc/lib/src/bloc_base.dart
felangel: Agreed but that would be breaking so I would propose differing that change until v9.0.0.

packages/bloc/lib/src/bloc_base.dart
felangel: Agreed but that would be breaking so I would propose differing that change until v9.0.0.

packages/bloc/lib/src/bloc_base.dart
felangel: It's technically breaking.

The following is currently allowed but with the change it would not compile:

```dart
class MyBloc extends Bloc<dynamic, dynamic> {...}
```

felangel: Yeah I can do `Object?` but `Object` would be breaking.

packages/flutter_bloc/lib/src/bloc_provider.dart
```diff
@@ -34,7 +34,7 @@ mixin BlocProviderSingleChildWidget on SingleChildWidget {}
 /// ```
 ///
 /// {@endtemplate}
-class BlocProvider<T extends BlocBase<Object?>>
+class BlocProvider<T extends StateStreamable<Object?>>
```
purplenoodlesoop, line 37: I am not very sure about this suggestion, but maybe be it will be appropriate to change constraints from `T extends StateStreamable<Object?>` to `T extends StateStreamable<Object?>, Closable`

felangel: If you're suggesting this syntax will imply that T extends StateStreamable and Closable then I don't think this will work.  As far as I know, Dart does not have intersection types.

narcodico: You can extend from a single class only indeed.

purplenoodlesoop: > If you're suggesting this syntax will imply that T extends StateStreamable and Closable then I don't think this will work. As far as I know, Dart does not have intersection types.

Oh, I'm so sorry for my wording and the misunderstanding caused! 

What I meant, is to create another interface that combines both `StateStreamable<Object?>` and `Closable` in one, something like

```dart
abstract class StateStreamableSource<T extends Object?> implements StateStreamable<T>, Closable { }
```

purplenoodlesoop (approved):
Awesome, thank you very much!

beeth0ven:
Hi there!

This is my 2 cents feedback about naming. It looks like `Stream` has an interface called [StreamView](https://github.com/dart-lang/sdk/blob/6091533c6e3f70969acdc65ac422298c05fa7567/sdk/lib/async/stream.dart#L1805) in `dart:async`. So we may consider names like `BlocView`.

And there is also a type called [ValueStream](https://github.com/ReactiveX/rxdart/blob/480b6de82dfdf938cc2ff6c5bd752acee3a6ac21/lib/src/streams/value_stream.dart#L2) in `rxdart` that expose getter for latest value and impelement `Stream`. Here are some options for our cases like `StateStream`, `CurrentStateStream` or `StateStreamView`. This type is combination of `Getter<State>` and `Stream<State>`.

Thanks!

beeth0ven:
>It looks like Stream has an interface called StreamView in dart:async. So we may consider names like BlocView.

After some research I found `*View` offten represent variant of original type like:
- [UnmodifiableListView](https://api.dart.dev/stable/2.14.4/dart-collection/UnmodifiableListView-class.html)
- [UnmodifiableSetView](https://api.dart.dev/stable/2.14.4/dart-collection/UnmodifiableSetView-class.html)
- [NonGrowableListView](https://pub.dev/documentation/collection/latest/collection/NonGrowableListView-class.html)

So `BlocView` may not be a good fit for our interface name. My point here is to find a better name than `Streamable`, I guess there will be one which is more friendly.

Thanks.

felangel:
> > It looks like Stream has an interface called StreamView in dart:async. So we may consider names like BlocView.
> 
> After some research I found `*View` offten represent variant of original type like:
> 
> * [UnmodifiableListView](https://api.dart.dev/stable/2.14.4/dart-collection/UnmodifiableListView-class.html)
> * [UnmodifiableSetView](https://api.dart.dev/stable/2.14.4/dart-collection/UnmodifiableSetView-class.html)
> * [NonGrowableListView](https://pub.dev/documentation/collection/latest/collection/NonGrowableListView-class.html)
> 
> So `BlocView` may not be a good fit for our interface name. My point here is to find a better name than `Streamable`, I guess there will be one which is more friendly.
> 
> Thanks.

Thanks for the feedback! Why do you feel streamable is a bad name? The reason for Streamable and StateStreamable is because they are analogous to Listenable and ValueListenable.

beeth0ven:
>Thanks for the feedback! Why do you feel Streamable is a bad name? The reason for Streamable and StateStreamable is because they are analogous to Listenable and ValueListenable.

Yeah, `Streamable` is not a bad name, in fact it's a good technical name. The bad feeling is directly form my heart 🙃. From my view point, names can help developer (library consumer) link the concept with previous experience, and get the bridge quickly. I've tried to search `Streamable` in my coding experience, but get empty response. `Streamable` make no sense for me. I may stand for someone that is not good at english ha.

Technical speaking.  I feel `Streamable` may express the interface of `Stream`:

```dart
abstract class Streamable<T> {
  StreamSubscription<T> listen(...);
  ... // other stream method
}
```

When we come to these types:

```dart
abstract class Type1<T> {
  Stream<T> get stream;
}
abstract class Type2<T> {
  T get state;
}
abstract class Type3<T> {
  T get state;
  Stream<T> get stream;
}
```

They looks like providers/holders/getters. I'd like to list some code to get the feeling:

1. Streamable:
```dart
abstract class Streamable<T> {
  Stream<T> get stream;
}
abstract class Stateble<T> {
  T get state;
}
abstract class StateStreamable<T> {
  T get state;
  Stream<T> get stream;
}

// consume this name
class BlocConsumer<B extends StateStreamable<S>, S> .... // suppose we don't know what is `StateStreamable`,
                                                         // can we guess what it is in 5 seconds
```

2. Provider/Holder:

```dart
abstract class StreamProvider<T> {
  Stream<T> get stream;
}
abstract class StateProvider<T> {
  T get state;
}
abstract class StateStreamProvider<T> {
  T get state;
  Stream<T> get stream;
}

// consume this name
class BlocConsumer<B extends StateStreamProvider<S>, S> .... // suppose we don't know what is `StateStreamProvider`,
                                                             // can we guess what it is in 5 seconds
```

3. Getters

```dart
abstract class HasStream<T> {
  Stream<T> get stream;
}
abstract class HasState<T> {
  T get state;
}
abstract class HasStateStream<T> {
  T get state;
  Stream<T> get stream;
}

// consume this name
class BlocConsumer<B extends HasStateStream<S>, S> .... // suppose we don't know what is `HasStateStream`,
                                                        // can we guess what it is in 5 seconds
```

not sure if it's just me feel the later one is a bit better. I think there is a name that may help us understand the concept more easily.

Thanks!

felangel:
> > Thanks for the feedback! Why do you feel Streamable is a bad name? The reason for Streamable and StateStreamable is because they are analogous to Listenable and ValueListenable.
> 
> Yeah, `Streamable` is not a bad name, in fact it's a good technical name. The bad feeling is directly form my heart 🙃. From my view point, names can help developer (library consumer) link the concept with previous experience, and get the bridge quickly. I've tried to search `Streamable` in my coding experience, but get empty response. `Streamable` make no sense for me. I may stand for someone that is not good at english ha.
> 
> Technical speaking. I feel `Streamable` may express the interface of `Stream`:
> 
> ```dart
> abstract class Streamable<T> {
>   StreamSubscription<T> listen(...);
>   ... // other stream method
> }
> ```
> 
> When we come to these types:
> 
> ```dart
> abstract class Type1<T> {
>   Stream<T> get stream;
> }
> abstract class Type2<T> {
>   T get state;
> }
> abstract class Type3<T> {
>   T get state;
>   Stream<T> get stream;
> }
> ```
> 
> They looks like providers/holders/getters. I'd like to list some code to get the feeling:
> 
> 1. Streamable:
> 
> ```dart
> abstract class Streamable<T> {
>   Stream<T> get stream;
> }
> abstract class Stateble<T> {
>   T get state;
> }
> abstract class StateStreamable<T> {
>   T get state;
>   Stream<T> get stream;
> }
> 
> // consume this name
> class BlocConsumer<B extends StateStreamable<S>, S> .... // suppose we don't know what is `StateStreamable`,
>                                                          // can we guess what it is in 5 seconds
> ```
> 
> 2. Provider/Holder:
> 
> ```dart
> abstract class StreamProvider<T> {
>   Stream<T> get stream;
> }
> abstract class StateProvider<T> {
>   T get state;
> }
> abstract class StateStreamProvider<T> {
>   T get state;
>   Stream<T> get stream;
> }
> 
> // consume this name
> class BlocConsumer<B extends StateStreamProvider<S>, S> .... // suppose we don't know what is `StateStreamProvider`,
>                                                              // can we guess what it is in 5 seconds
> ```
> 
> 3. Getters
> 
> ```dart
> abstract class HasStream<T> {
>   Stream<T> get stream;
> }
> abstract class HasState<T> {
>   T get state;
> }
> abstract class HasStateStream<T> {
>   T get state;
>   Stream<T> get stream;
> }
> 
> // consume this name
> class BlocConsumer<B extends HasStateStream<S>, S> .... // suppose we don't know what is `HasStateStream`,
>                                                         // can we guess what it is in 5 seconds
> ```
> 
> not sure if it's just me feel the later one is a bit better. I think there is a name that may help us understand the concept more easily.
> 
> Thanks!

Really appreciate the feedback and detailed response! The main issue with something like `StateProvider` or `StateStreamProvider` is it will conflict with the existing notion of `Providers` (from `package:provider` or `package:riverpod`).

Regarding the latter, `HasStateStream`, `HasState`, etc. while they are descriptive, I don't think they align with any other naming conventions in Dart/Flutter for similar interfaces.

To the point about being able to guess what the interface is within 5 seconds, I feel `Streamable` implies something that exposes a `stream` which can be listened to (but maybe it's just me 😅). I also think naming is really hard and ultimately the API documentation is there to make it crystal clear what the interface means. 

I'd love to hear what others think/feel about the naming. I can try to put out a poll to get more feedback from the community 👍

purplenoodlesoop:
I fully agree with @felangel that `StreamProvider` and `HasStream` are not very good. `Streamable` is fine, and it does go pretty well with the existing `Listenable`, but IMO its meaning is not quite right.

How I see it, is that `Listenable` is something that can be listened to, thus, `Streamable` is something that can be streamed, and by "something" I mean the class itself. So some class that implements the `Streamable` interface sounds like some entity that can be streamed itself, if that makes sense?

I see a `StreamListenable` as a good alternative.

packages/flutter_bloc/lib/src/bloc_builder.dart
```diff
@@ -69,7 +69,8 @@ typedef BlocBuilderCondition<S> = bool Function(S previous, S current);
 /// )
 /// ```
 /// {@endtemplate}
-class BlocBuilder<B extends BlocBase<S>, S> extends BlocBuilderBase<B, S> {
+class BlocBuilder<B extends StateStreamable<S>, S>
```
lhmzhou, line 72: Why not add null safety checks here?

```
class BlocBuilder<B? extends StateStreamable<S>, S?>
```

felangel: Because the bloc should not be null

beeth0ven:
Thanks for the reply.

I have been a bit busy recently, and I may not be fully awake yet, but I don't want to delay everyone's time. Please forgive me if the following replies have improper words 🙏.

from felangel:
>Really appreciate the feedback and detailed response! The main issue with something like StateProvider or StateStreamProvider is it will conflict with the existing notion of Providers (from package:provider or package:riverpod).

I totaly agreed. This is what I am worried about if we use `Providers`, since developer need to tell which `StreamProvider` is it, they may have mix feeling.

>Regarding the latter, HasStateStream, HasState, etc. while they are descriptive, I don't think they align with any other naming conventions in Dart/Flutter for similar interfaces.

Yeah, these names are not commonly used, I only found them in a swift library `NSObject-Rx` that introduce [HasDisposeBag](https://github.com/RxSwiftCommunity/NSObject-Rx/blob/91ad389faf34ea0daaae188bdfeae54150f1cbf0/HasDisposeBag.swift#L8).

>To the point about being able to guess what the interface is within 5 seconds, I feel Streamable implies something that exposes a stream which can be listened to (but maybe it's just me 😅). I also think naming is really hard and ultimately the API documentation is there to make it crystal clear what the interface means.

>I'd love to hear what others think/feel about the naming. I can try to put out a poll to get more feedback from the community 👍

em, I'd love to listen to the community too ha. About the guess work (guess within 5 seconds), it is a way I used to test if the name can be widely accept. since there seem no objective way to tell if a name is good or not that good (production manager sometime use similar way to verify earily ideas with users).

from purplenoodlesoop:
>How I see it, is that Listenable is something that can be listened to, thus, Streamable is something that can be streamed, and by "something" I mean the class itself. So some class that implements the Streamable interface sounds like some entity that can be streamed itself, if that makes sense?

yes, I'm expressing same thing as you.

We are all comfortable with these interface:

```dart
abstract class  Listenable {
  void listen(...);
}
abstract class Cancelable {
  void cancel();
}
abstract class Closable {
  void close();
}
```

If I understand this proposal correctly. what we are achieving is split object into piece of abstraction/trait:

Before:

```dart
// bellow code are not strictly correct, thay are just for concept demo
class BlocBase<State> {
  State get state => ...;
  Stream<State> get stream => ...;
  FutureOr<void> clsose() { ... }
  ... // other methods
}

class BlocBuilderBase<B extends BlocBase<S>, S> { // A widget that consume bloc
  // in the implemetation detail, we only need to call 3 methods from bloc:
  // bloc.state, bloc.stream and bloc.close().
  // so this widget just need a class provides these 3 methods ,
  // we can extract an interface for this, then we can reuse this widget with other bloc implementation
 ...
} 
...
```

After:

```dart
abstract class Streamable<S> {
  Stream<S> get stream;
}
abstract class StateStreamable<S> implements Streamable<S> {
  State get state;
}
abstract class Closable {
  FutureOr<void> close();
}
abstract class StateStreamableSource<S>
  implements StateStreamable<S>, Closable {
  // this interface provides methods `BlocBuilderBase` needs as:
  // State get state;
  // Stream<S> get stream;
  // FutureOr<void> close();
} 

class BlocBuilderBase<B extends StateStreamableSource<S>, S> { 
  // `BlocBuilderBase` can be reused if that object 
  // provide these 3 methods (implements StateStreamableSource)
 ...
} 
```

The problem I see is the classification of these interfaces:

```dart
abstract class Streamable<S> {
  Stream<S> get stream;
}
abstract class Stateble<T> {
  T get state;
}
abstract class StateStreamable<S> {
  State get state;
  Stream<S> get stream;
}
```

If we follow the `*able` naming convention,  there is a certain probability it's hard to find a name for some types:

```dart
abstract class Type1 {
  User? get author;
}
abstract class Type2 {
  String? get title;
}
abstract class Type3 {
  Datetime? get creationDate;
}
```

Not sure if it's too earily or unrealistic to discuss these, since when we follow this pattern further, we may face this issue. If we have a clear way to name getters from the start, then this is not a problem.

I feel I'm a bit boring 😄. Ignore me if you are not comfortable with above statement.

I'm okay to follow the community. I'm fine with the community's choice. I just give feedback since there is a `feedback wanted` label in this PR.

Thank you.

felangel:
> I fully agree with @felangel that `StreamProvider` and `HasStream` are not very good. `Streamable` is fine, and it does go pretty well with the existing `Listenable`, but IMO its meaning is not quite right.
> 
> How I see it, is that `Listenable` is something that can be listened to, thus, `Streamable` is something that can be streamed, and by "something" I mean the class itself. So some class that implements the `Streamable` interface sounds like some entity that can be streamed itself, if that makes sense?
> 
> I see a `StreamListenable` as a good alternative.

I don't fully agree. A class that itself can be streamed is a `Stream` 😅

If we go with `StreamListenable` then my immediate thought would be that it implements `Listenable` which is not the case.

beeth0ven:
Hi there! 

I tried to find synonyms of provider then come with supplier:

```dart
abstract class StreamSupplier<T> {
  Stream<T> get stream;
}
abstract class StateSupplier<T> {
  T get state;
}
abstract class StateStreamSupplier<T> {
  T get state;
  Stream<T> get stream;
}

// consume this name
class BlocConsumer<B extends StateStreamSupplier<S>, S> ..
```

It is compatible with other types:

```dart
abstract class AuthorSupplier {
  User? get author;
}
abstract class TitleSupplier {
  String? get title;
}
abstract class CreationDateSupplier {
  Datetime? get creationDate;
}
```

If it make sense to us, what do you think about it 😊?

felangel:
> Hi there!
> 
> I tried to find synonyms of provider then come with supplier:
> 
> ```dart
> abstract class StreamSupplier<T> {
>   Stream<T> get stream;
> }
> abstract class StateSupplier<T> {
>   T get state;
> }
> abstract class StateStreamSupplier<T> {
>   T get state;
>   Stream<T> get stream;
> }
> 
> // consume this name
> class BlocConsumer<B extends StateStreamSupplier<S>, S> ..
> ```
> 
> It is compatible with other types:
> 
> ```dart
> abstract class AuthorSupplier {
>   User? get author;
> }
> abstract class TitleSupplier {
>   String? get title;
> }
> abstract class CreationDateSupplier {
>   Datetime? get creationDate;
> }
> ```
> 
> This naming is pretty straightforward, what do you think about it 😊?

Thanks for the additional feedback. I still have reservations about that naming convention because as far as I can tell `Supplier` is not a standard naming convention for similar interfaces (particularly not within Flutter and Dart). Whereas `Streamable` and `StateStreamable` is modeled after `Listenable` and `ValueListenable`.

packages/bloc/lib/src/bloc.dart
```diff
@@ -7,6 +7,16 @@ part 'bloc_base.dart';
 part 'bloc_overrides.dart';
 part 'emitter.dart';
 
+/// An [ErrorSink] that supports adding events.
+///
+/// Multiple events can be reported to the sink via `add`.
+abstract class EventSink<Event extends Object?> implements ErrorSink {
```
purplenoodlesoop, line 13: [EventSink is an existing interface](https://api.dart.dev/stable/2.14.4/dart-async/EventSink-class.html) from the `dart:async` library

felangel: Yup I went back and forth on it but ultimately decided it made more sense to have a custom EventSink so that it implements ErrorSink rather than two unrelated classes that happen to have the same API. Also if we go back to implementing EventSink from dart:async we’ll be prone to the false positives with the close_sinks lint rule.

purplenoodlesoop: > it made more sense to have a custom EventSink
I totally agree! One thing is I am a bit worried about is the name clash :) Currently was testing this branch and had to revert to named imports in order to use both EventSinks.

felangel: > I totally agree! One thing is I am a bit worried about is the name clash :) Currently was testing this branch and had to revert to named imports in order to use both EventSinks.

Yes, if you want to use both you’d need named imports but is there a common use case in which you’d need to import both interfaces within the same file?

> Also, having two classes with the same name and relatively the same meaning from two different libraries can be confusing.

Agreed, but I think it’d be more confusing to have Bloc implement both EventSink and ErrorSink where the two come from different libraries but have the same API.

felangel: @purplenoodlesoop I suppose we can rename Sink to something else like Pipe, Vent, or Chute 🤷‍♂️

What do you think?

purplenoodlesoop: > Agreed, but I think it’d be more confusing to have Bloc implement both EventSink and ErrorSink where the two come from different libraries but have the same API.

100%, I agree that a custom interface is needed.

> I suppose we can rename Sink to something else like Channel, Consumer, Receiver, or Pipe 

What about `BlocEventSink`?

felangel: Yeah that works as well. My only concern is from reading the class name I’d assume it implements EventSink from dart:async 😅

purplenoodlesoop: Whoops, indeed. Namings are hard 🥲

felangel: Yeah I know I’m terrible at naming 😅

I will update to BlocEventSink for now though since I agree that’s a better solution.

purplenoodlesoop: > Yeah I know I’m terrible at naming 😅

No no, namings are indeed tricky :) Or we all are terrible at them lol

beeth0ven:
>Thanks for the additional feedback. I still have reservations about that naming convention because as far as I can tell Supplier is not a standard naming convention for similar interfaces (particularly not within Flutter and Dart). Whereas Streamable and StateStreamable is modeled after Listenable and ValueListenable.

Thanks for the update.

Yeah, I agreed naming is hard 😄. I may be too obsessed with naming. Sorry for that. 

Since `Streamable` best fit the current dart naming convention, I'm fine with it now.

Thank @felangel, @purplenoodlesoop, @PlugFox , @lhmzhou for taking time to discuss and improve this library!

narcodico:
I'd like to point out that with this approach we won't be able to `add` events by fetching an existing instance through `context`, e.g.: `context.read<CounterCubit>().add(...)`, since `StateStreamableSource` doesn't expose `add`.

felangel:
> I'd like to point out that with this approach we won't be able to `add` events by fetching an existing instance through `context`, e.g.: `context.read<CounterCubit>().add(...)`, since `StateStreamableSource` doesn't expose `add`.
> 
> This would go against the bloc pattern which relies on the user being able to add events from the presentation layer to the sink.

It should still work. The generic is still of type T where T extend StateStreamableSource but as a developer when you pass BlocProvider.of<MyBloc>(context); you’ll get an instance of MyBloc back so nothing should change from the user’s perspective.

Let me know if I missed something, thanks!

narcodico:
I missed something, you can discard what I said.

jorgecoca (approved):

PR #3042, opened 2021-12-04
feat(cubit): optional StateError when emit happens

### LINKED ISSUE #2980: question: StateError
**Description**
In 8.0.0 if bloc is closed both `emit` and `add` throws `StateError`. In Merge Request is mentioned [2261](https://github.com/felangel/bloc/issues/2261) which refers to [1236](https://github.com/felangel/bloc/issues/1236) and [2912](https://github.com/felangel/bloc/pull/2912). I'm on the side of last two, because most happening case is `Future` resolved after bloc is closed.

For example is started inside bloc, user goes away from the screen before `Future` is resolved and bloc is closed now. When `Future` is resolved it calls emit which results in `StateError`.

So my questions how to properly handle this case, because `Future`s cannot be canceled?
Just adding if `(isClosed)` before each of my emit in every bloc seem's like a lot of code, and overriding `emit` is not an options since it's an argument of event handler

**Steps To Reproduce**

1. Go to Page with Bloc on it
2. Trigger event with Future in it
3. Leave the Page before Future is resolved
4. See StateError

**Additional Context**
I've started using BLoC instead of Provider because Provider was getting error about it being updated after dispose.

maRci002:
## Status

**READY**

## Breaking Changes

NO

## Description

This PR focuses on Local Cubits where the cubit might become closed / disposed during processing async Cubit functions.

Current problem and workarounds:
```dart
class SomeCubit extends Cubit<int> {
  SomeCubit() : super(0);

  Future<void> test() async {
    try {
      emit(1);
      await Future.delayed(const Duration(seconds: 5));
      // if user leaves the page / dialog E.g. `pushReplacementNamed` then cubit is already closed
      // check `isClosed` every time after a future completes?
      if (!isClosed) {
        emit(2);
      }
    } on StateError {
      // here we could ignore StateError('Cannot emit new states after calling close')
      // but this seems overkill for every single cubit function
    }
  }
}
```

Syntax if programmer knows the Cubit is a local cubit / temporary cubit:
```dart
SomeCubit() : super(0, emitFailsWhenClosed: false)
```
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

closes #2980

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -38,10 +38,7 @@ class _Handler {
 /// {@endtemplate}
 abstract class Bloc<Event, State> extends BlocBase<State> {
   /// {@macro bloc}
-  Bloc(
-    State initialState, {
-    bool emitFailsWhenClosed = true,
-  }) : super(initialState, emitFailsWhenClosed: emitFailsWhenClosed);
+  Bloc(State initialState) : super(initialState, emitFailsWhenClosed: true);
 
   final _eventController = StreamController<Event>.broadcast();
   final _subscriptions = <StreamSubscription<dynamic>>[];
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -7,6 +7,16 @@ part 'bloc_base.dart';
 part 'bloc_overrides.dart';
 part 'emitter.dart';
 
+/// An [ErrorSink] that supports adding events.
+///
+/// Multiple events can be reported to the sink via `add`.
+abstract class BlocEventSink<Event extends Object?> implements ErrorSink {
+  /// Adds an [event] to the sink.
+  ///
+  /// Must not be called on a closed sink.
+  void add(Event event);
+}
+
 /// An event handler is responsible for reacting to an incoming [Event]
 /// and can emit zero or more states via the [Emitter].
 typedef EventHandler<Event, State> = FutureOr<void> Function(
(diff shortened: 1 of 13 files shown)


maRci002:
I'm sorry I need to do a little refactor since bloc already [awaits ongoing event emitters](https://github.com/felangel/bloc/blob/bloc-v8.0.1/packages/bloc/lib/src/bloc.dart#L520-L535).

> In addition, if [close] is called while `events` are still being processed, the [Bloc] will finish processing the pending `events`.

However Cubit doesn't have this feature so I'm going to apply the changes only for it.

lhmzhou (approved):

lhmzhou (approved):

maRci002:
@felangel will you please review this PR?

Maybe `Cubit` should give explicit `emitFailsWhenClosed: false` to his base contructor (so all Cubits are safe) or `BlocBase` should mark `bool get emitFailsWhenClosed` accessor as abstract.
Or just completly refactor Cubit's logic for instance keep track of async Cubit functions.

felangel:
> @felangel will you please review this PR?
> 
> Maybe `Cubit` should give explicit `emitFailsWhenClosed: false` to his base contructor (so all Cubits are safe) or `BlocBase` should mark `bool get emitFailsWhenClosed` accessor as abstract. Or just completly refactor Cubit's logic for instance keep track of async Cubit functions.

I'd prefer to hold off on this change for a little bit and evaluate other potential solutions. I want to make sure we carefully consider all options/approaches to resolve this before adding a new public API. Really appreciate you taking the time to create this PR 👍 

I'll think about it a bit more over the next few days and get back to you, thanks! 🙏

felangel:
@maRci002 I'd love your thoughts on https://github.com/felangel/bloc/issues/3069, thanks! 🙏

felangel:
Closing this for now since we need to discuss/align on an approach first (#3069)

PR #3066, opened 2021-12-15
refactor(bloc): add isClosed to Closable interface

### LINKED ISSUE #3065: refactor: further apply the DIP and ISP principles
Hello!

Thank you so much for the recent "core interfaces" update! It is really a lifesaver! 

I've identified three places where it can be extended to further apply the same principles.

1. Extract "hooks" to separate interfaces. The `onChange` and `onError` can be extracted in one interface, `BlocBaseHooks`, and the `onEvent` and `onTransition` can be extracted to the second, `BlocHooks`, that will implement the first.

2. The `isClosed` getter can be moved to the `Closable` interface.

3. `BlocObserver` can depend on interfaces using the following schema with explicit covariance.

|  Method  |  New type  |
|---|---|
|  `onCreate`  |  `StateStreamableSource`  |
|  `onEvent`  |  `BlocEventSink`  |
|  `onChange`  |  `StateStreamable`  |
|  `onTransition`  |  `BlocEventSink`  |
|  `onError`  |  `ErrorSink`  |
|  `onClose`  |   `Closable`  |

I'm almost done with the PR and looking forward to seeing what you think :)

purplenoodlesoop:
Closes #3065

## Status

**READY**

## Breaking Changes

NO

## Description

There are three main changes done in the following PR. All of them are targeted to extend the recent "core interfaces" update to further apply the Interfaces Segregation and Dependency Inversion principles.

Refer to #3065

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/bloc_base.dart
+++ b/packages/bloc/lib/src/bloc_base.dart
@@ -21,6 +21,11 @@ abstract class Closable {
   /// Closes the current instance.
   /// The returned future completes when the instance has been closed.
   FutureOr<void> close();
+
+  /// Whether the object is closed.
+  ///
+  /// An object is considered closed once [close] is called.
+  bool get isClosed;
 }
 
 /// An object that can emit new states.
@@ -72,6 +77,7 @@ abstract class BlocBase<State>
   ///
   /// A bloc is considered closed once [close] is called.
   /// Subsequent state changes cannot occur within a closed bloc.
+  @override
   bool get isClosed => _stateController.isClosed;
 
   /// Updates the [state] to the provided [state].
@@ -122,6 +128,7 @@ abstract class BlocBase<State>
   ///
   /// * [BlocObserver] for observing [Cubit] behavior globally.
   ///
+  @protected
   @mustCallSuper
   void onChange(Change<State> change) {
     // ignore: invalid_use_of_protected_member
@@ -129,6 +136,7 @@ abstract class BlocBase<State>
   }
 
   /// Reports an [error] which triggers [onError] with an optional [StackTrace].
+  @protected
   @mustCallSuper
   @override
   void addError(Object error, [StackTrace? stackTrace]) {
--- a/extensions/intellij/README.md
+++ b/extensions/intellij/README.md
@@ -58,6 +58,9 @@ Right click and use `Generate -> Equatable Props` to automatically generate the 
 | `blocstate`         | Creates a state class                           |
 | `blocevent`         | Creates an event class                          |
 | `bloctest`          | Creates a `blocTest` with build, act and expect |
+| `mockbloc`          | Creates a class extenting `MockBloc`            |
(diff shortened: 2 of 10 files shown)


packages/bloc/lib/src/bloc_base.dart
```diff
@@ -39,12 +44,57 @@ abstract class ErrorSink implements Closable {
   void addError(Object error, [StackTrace? stackTrace]);
 }
 
+/// An interface for hooks of the [BlocBase].
+///
+/// Contains the `onChange` and `onError` hooks.
+abstract class BlocBaseHooks<State> {
```
felangel, line 50: thoughts on `LifecycleObserver` and `BlocLifecycleObserver`? I'm not a fan of `hooks` because the term hooks is typically associated with `flutter_hooks`.

packages/bloc/lib/src/bloc_base.dart
```diff
@@ -21,6 +21,11 @@ abstract class Closable {
   /// Closes the current instance.
   /// The returned future completes when the instance has been closed.
   FutureOr<void> close();
+
+  /// Whether the [Closable] is closed.
+  ///
+  /// A [Closable] is considered closed once [close] is called.
+  bool get isClosed;
```
felangel, line 28: 🚀

packages/bloc/lib/src/bloc_base.dart
felangel, line 50: What's the benefit of introducing this interface? I don't like the name `hooks` because it can be confused with `flutter_hooks` and I also would prefer not to define interfaces unnecessarily. Is there a concrete use-case you had in mind which required operating on this interface specifically?

purplenoodlesoop: > What's the benefit of introducing this interface?

The main benefits align with those that ISP and DIP brings — other object can implement the lifecycle calls more easily, other object interested only in lifecycle calls can depend only on desired parts, making code more loosely coupled. 

> I don't like the name hooks because it can be confused with flutter_hooks and I also would prefer not to define interfaces unnecessarily. 

Totally agree, the name can be confusing. `LifecycleObserver` sounds good!

> Is there a concrete use-case you had in mind which required operating on this interface specifically?

Yes, I was implementing a custom, project-specific version of Bloc that needed some of the "on.." methods, and had to either just hard-code them in the implementation, or extend/implement the BlocBase, but not all methods was desired. 

Those interfaces would allow to unify lifecycle calls. 

On the other hand, I agree that non-essential interfaces can be redundant and open to discard them if you think that it's too much :)

> Also, what constitutes a hook? I would expect onCreate, onClose, etc. to also be included.

I agree, they should be also included if we decide to keep the interfaces!

felangel (approved):

PR #3070, opened 2021-12-16
fix(bloc_test): throw uncaught exceptions

### LINKED ISSUE #3058: fix: [bloc_test] Incorrect error message when exception occurs in "act"
**Description**
When an exception occurs in the "act" phase of blocTest, the exception isn't correct/helpful in actually locating the issue. This is because blocTest assumes only the bloc can throw exceptions (and handles those), but if an error occurs in for example in an event constructor, this is not displayed to the user.

**Steps To Reproduce**
Minimal repro case to reproduce:
```dart

void main() {
  blocTest<TestBloc, TestState>(
    "Test",
    build: () => TestBloc(TestState()),
    act: (bloc) => bloc.add(TestEvent(TestEntity())),
    expect: () => [isA<TestState>()],
  );
}

class TestBloc extends Bloc<TestEvent, TestState> {
  TestBloc(TestState initialState) : super(initialState) {
    on<TestEvent>((event, emit) => emit(TestState()));
  }
}

class TestEvent {
  TestEntity entity;
  TestEvent(this.entity);
}

class TestState {}

class BaseEntity {
  String name;
  BaseEntity({required this.name});
}

class TestEntity extends BaseEntity {
  TestEntity({name}) : super(name: name);
}
```
**Current Behavior**
```
Expected: [<<Instance of 'TestState'>>]
  Actual: []
   Which: at location [0] is [] which shorter than expected
```

**Expected Behavior**
I would expect blocTest to display the correct exception (and stacktrace!), being: 
```
Unhandled exception:
type 'Null' is not a subtype of type 'String'
```

**Additional Context**
We had this error occur in our code base. I had incorrectly assumed the type of the constructor argument of TestEntity was type inferred from BaseEntity, but I didn't notice it's marked as dynamic when you hover over it. This doesn't cause any issues until someone omits the argument, which a colleague did in a test case, which sent him down a wild chase to find what was wrong with his test/use of blocTest.

Thanks for the work on bloc/bloc_test! 👍

felangel:
## Status

**READY**

## Breaking Changes

NO

## Description

If an uncaught error occurs within a bloc, cubit, or within the `blocTest` they are currently being swallowed by default which makes it very difficult to debug. Instead of swallowing uncaught errors, this change rethrows uncaught errors so that it is clear that an exception has occurred during the `blocTest`.

If `errors` is not null, uncaught errors will be aggregated and evaluated against the `errors` matcher.

- closes #3058

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -200,6 +200,7 @@ Future<void> testBloc<B extends BlocBase<State>, State>({
           try {
             await act?.call(bloc);
           } catch (error) {
+            if (errors == null) rethrow;
             unhandledErrors.add(error);
           }
           if (wait != null) await Future<void>.delayed(wait);
@@ -231,7 +232,7 @@ WARNING: Please ensure state instances extend Equatable, override == and hashCod
 Alternatively, consider using Matchers in the expect of the blocTest rather than concrete state instances.\n''',
             );
           }
-          if (!unhandledErrors.contains(error)) {
+          if (errors == null || !unhandledErrors.contains(error)) {
             // ignore: only_throw_errors
             throw error;
           }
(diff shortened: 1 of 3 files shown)


felangel:
cc @orestesgaolin

definitelyokay (approved):

jeroen-meijer (approved):
Noice :shipit:

orestesgaolin:
👍

PR #3167, opened 2022-01-24
docs(bloc): Included missing slash in doc comment

### LINKED ISSUE #3166: docs: Include missing slash in doc comment
**Description**

I think line 34 of `emitter.dart` (this is the first line of the method documentation for `forEach`)  is missing an `/`, I believe it should be `///`.  In other words I believe the block comment should be a doc comment.

This is:
```dart
  // Subscribes to the provided [stream] and invokes the [onData] callback
  /// when the [stream] emits new data and the result of [onData] is emitted.
  ///
  /// [forEach] completes when the event handler is cancelled or when
  /// the provided [stream] has ended.
  ///
  /// If [onError] is omitted, any errors on this [stream]
  /// are considered unhandled, and will be thrown by [forEach].
  /// As a result, the internal subscription to the [stream] will be canceled.
  ///
  /// If [onError] is provided, any errors on this [stream] will be passed on to
  /// [onError] and will not result in unhandled exceptions or cancelations to
  /// the internal stream subscription.
  ///
  /// **Note**: The stack trace argument may be [StackTrace.empty]
  /// if the [stream] received an error without a stack trace.
  Future<void> forEach<T>(
    Stream<T> stream, {
    required State Function(T data) onData,
    State Function(Object error, StackTrace stackTrace)? onError,
  });
```

I'm aware that according to the [Flutter Style Guide](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#use-for-public-quality-private-documentation) "Documentation of private APIs that is not of sufficient quality should only use //". However, by reading this, I think line 34 might be just missing an extra `/`. Since it seems it does have sufficient quality.

 In addition, the `onEach` method documentation begins with a similar phrase and it is using a `///` instead of `//`. 
 
 Hence,  I think an additional `/` should be added in line 34.

alestiago:
## Status

**READY**

## Breaking Changes

NO

## Description

Included an additional slash to convert the block comment (`//`) to a doc comment (`///`). Closes #3166 .

## Type of Change

- [ ] 🧹 Code refactor
- [X] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/emitter.dart
+++ b/packages/bloc/lib/src/emitter.dart
@@ -31,7 +31,7 @@ abstract class Emitter<State> {
     void Function(Object error, StackTrace stackTrace)? onError,
   });
 
-  // Subscribes to the provided [stream] and invokes the [onData] callback
+  /// Subscribes to the provided [stream] and invokes the [onData] callback
   /// when the [stream] emits new data and the result of [onData] is emitted.
   ///
   /// [forEach] completes when the event handler is cancelled or when


felangel (approved):
Thanks for the contribution! 💙🎉

Issue #1752, opened 2020-09-21
[flutter_bloc] Proposal: add child parameter to BlocBuilder constructor


iapicca:
Adding a child parameter to BlocBuilder could improve efficiency and reduce boilerplate,
especially in cases where the bloc affect a widget but not its child, as for the example below.

In the current situation you have either:

1. Pointless rebuild

```dart
@override
 Widget build(BuildContext context) => BlocBuilder<OpacityCubit, double>(
    builder: (BuildContext context, double state) => Opacity(
    opacity: state,
    child: SomeWidget(), // <= rebuilds pointlessly
  ),
);
```

2. boilerplate and limitations 

```dart
 final Widget _child = SomeWidget();  // <= could be a problem in stateless widget with const constructor

 @override
 Widget build(BuildContext context) => BlocBuilder<OpacityCubit, double>(
   builder: (BuildContext context, double state) => Opacity(
    opacity: state,
    child: _child,
  ),
);
```

```dart
 @override
 Widget build(BuildContext context) {
   final Widget _child = SomeWidget();  // not fond of declaring stuff in the build function if I can avoid it
   return BlocBuilder<OpacityCubit, double>(
     builder: (BuildContext context, double state) => Opacity(
       opacity: state,
       child: _child,
    ),
  );
}
```

I propose something like the example below (an [approach](https://youtu.be/6KiPEqzJIKQ) similar to some implicit animation)

```dart
Widget build(BuildContext context) => BlocBuilder<OpacityCubit, double>(
  child: SomeWidget(),
  builder: (BuildContext context, double state) => Opacity(
    opacity: state,
    child: child,
  ),
);
```

iapicca:
@felangel sorry I didn't see this [issue](https://github.com/felangel/bloc/issues/1596) I noticed only from the [PR](https://github.com/felangel/bloc/pull/1597) you linked
I won't take it personal if this issue get closed as duplicate

felangel:
@iapicca the main problem I have with this feature request is justifying the breaking, verbose API will be worthwhile when in most cases this is an extra optimization. In my opinion, having a child and builder is confusing and adds noise to the API. Also, I think it's worthwhile pointing out that `StreamBuilder` does not currently offer support for this when the same argument can be made for including it -- it'd be interesting to raise this issue on Flutter in regards to `StreamBuilder` to see if there's a reason why support for `child` hasn't been added to `StreamBuilder`. Thoughts?

iapicca:
[...]

PR #3290, opened 2022-03-22
feat(flutter_bloc): Add child constructor to BlocBuilder, BlocConsumer and BlocSelector

mahdi739:
## Status

**READY**

## Breaking Changes

NO

## Description

It adds a named constructor to BlocBuilder, BlocConsumer and BlocSelector called "child".
This constructor has an extra parameter called "child" which is a widget that won't be rebuilt during state changes. 
Also, in this constructor you need to use "builderChild" instead of "builder" function which provides you that pre-built child to place in widget-tree.

It's good for when a part of widget tree doesn't depend on the state and can't be created outside of the BlocBuilder. (Like when you want to change color of container but not its child)
The good point is that we don't have to refactor our old code and we can use the default constructor as well.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_builder.dart
+++ b/packages/flutter_bloc/lib/src/bloc_builder.dart
@@ -6,6 +6,14 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 /// This is analogous to the `builder` function in [StreamBuilder].
 typedef BlocWidgetBuilder<S> = Widget Function(BuildContext context, S state);
 
+/// Signature for the `builderChild` function which takes the `BuildContext`
+/// ,[child] and [state] and is responsible for returning a widget which is to
+/// be rendered. The child should typically be part of the returned widget tree.
+/// This is analogous to the `builder` function in [StreamBuilder] but with
+/// a [child] parameter wich is built only the first time.
+typedef BlocWidgetBuilderChild<S> = Widget Function(
+    BuildContext context, S state, Widget child);
+
 /// Signature for the `buildWhen` function which takes the previous `state` and
 /// the current `state` and is responsible for returning a [bool] which
 /// determines whether to rebuild [BlocBuilder] with the current `state`.
@@ -75,19 +83,46 @@ class BlocBuilder<B extends StateStreamable<S>, S>
   /// {@macro bloc_builder_build_when}
   const BlocBuilder({
     Key? key,
-    required this.builder,
+    required BlocWidgetBuilder<S> builder,
     B? bloc,
     BlocBuilderCondition<S>? buildWhen,
-  }) : super(key: key, bloc: bloc, buildWhen: buildWhen);
+  })  : builder = builder,
+        builderChild = null,
+        super(key: key, bloc: bloc, buildWhen: buildWhen, child: null);
+
+  /// Similar to [BlocBuilder] but gives a [child] parameter
+  /// as a pre-built subtree and pass it back to the `builderChild` function.
+  /// The [child] is built only the first time. This is good for whene
+  /// a part of your widget doesn't depend on the state, so you want
+  /// to prevent it from rebuilding every time the state changes.
+  const BlocBuilder.child({
+    required Widget child,
+    Key? key,
+    required BlocWidgetBuilderChild<S> builderChild,
+    B? bloc,
+    BlocBuilderCondition<S>? buildWhen,
+  })  : builder = null,
+        builderChild = builderChild,
+        super(key: key, bloc: bloc, buildWhen: buildWhen, child: child);
 
   /// The [builder] function which will be invoked on each widget build.
   /// The [builder] takes the `BuildContext` and current `state` and
   /// must return a widget.
   /// This is analogous to the [builder] function in [StreamBuilder].
-  final BlocWidgetBuilder<S> builder;
+  final BlocWidgetBuilder<S>? builder;
+
+  /// The [builderChild] function which will be invoked on each widget build.
+  /// The [builderChild] takes the `BuildContext` and current `state` and
+  /// `child` and must return a widget. The child should typically be part
+  /// of the returned widget tree.
+  /// This is analogous to the [builder] function in [StreamBuilder] but with
+  /// a [child] parameter wich is built only the first time.
+  final BlocWidgetBuilderChild<S>? builderChild;
 
   @override
-  Widget build(BuildContext context, S state) => builder(context, state);
+  Widget build(BuildContext context, S state, Widget? child) =>
+      builder?.call(context, state) ??
+      builderChild!.call(context, state, child!);
 }
 
 /// {@template bloc_builder_base}
@@ -101,7 +136,8 @@ class BlocBuilder<B extends StateStreamable<S>, S>
 abstract class BlocBuilderBase<B extends StateStreamable<S>, S>
     extends StatefulWidget {
   /// {@macro bloc_builder_base}
-  const BlocBuilderBase({Key? key, this.bloc, this.buildWhen})
+  const BlocBuilderBase(
+      {Key? key, this.bloc, this.buildWhen, required this.child})
       : super(key: key);
 
   /// The [bloc] that the [BlocBuilderBase] will interact with.
@@ -109,11 +145,16 @@ abstract class BlocBuilderBase<B extends StateStreamable<S>, S>
   /// [BlocProvider] and the current `BuildContext`.
   final B? bloc;
 
+  /// If the pre-built subtree is passed as the [child] parameter, the
+  /// [BlocBuilder] will pass it back to the `builderChild` function so that it
+  /// can be incorporated into the build.
+  final Widget? child;
+
   /// {@macro bloc_builder_build_when}
   final BlocBuilderCondition<S>? buildWhen;
 
   /// Returns a widget based on the `BuildContext` and current [state].
-  Widget build(BuildContext context, S state);
+  Widget build(BuildContext context, S state, Widget? child);
 
   @override
   State<BlocBuilderBase<B, S>> createState() => _BlocBuilderBaseState<B, S>();
@@ -163,7 +204,7 @@ class _BlocBuilderBaseState<B extends StateStreamable<S>, S>
       bloc: _bloc,
       listenWhen: widget.buildWhen,
       listener: (context, state) => setState(() => _state = state),
-      child: widget.build(context, _state),
+      child: widget.build(context, _state, widget.child),
     );
   }
 }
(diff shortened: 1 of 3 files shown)


mahdi739:
@felangel 
Hello Mr. Angelov 👋
Yes it's pretty similar and both have same purpose. 
But the main difference is that I created a named constructor (instead of modifying default constructor) so there's no need to refactoring old codes. No one have to even know about this change, just whoever need this option can use it.

felangel:
> @felangel Hello Mr. Angelov 👋 Yes it's pretty similar and both have same purpose. But the main difference is that I created a named constructor (instead of modifying default constructor) so there's no need to refactoring old codes. No one have to even know about this change, just whoever need this option can use it.

Can you provide a real-world example of when you'd use `child`? I'm not sure I understand when this type of optimization would be beneficial. It makes sense for widgets like `AnimatedBuilder` because you are dealing with animations but I'm not convinced it makes sense in the context of bloc. In addition, you can always create your own widget that wraps `BlocBuilder` to apply this change and I'd prefer not to include this unless there's a very strong argument in its favor. Let me know what you think, thanks!

mahdi739:
> 
> Can you provide a real-world example of when you'd use `child`? I'm not sure I understand when this type of optimization would be beneficial. It makes sense for widgets like `AnimatedBuilder` because you are dealing with animations but I'm not convinced it makes sense in the context of bloc. In addition, you can always create your own widget that wraps `BlocBuilder` to apply this change and I'd prefer not to include this unless there's a very strong argument in its favor. Let me know what you think, thanks!

In flutter most of things are widgets so they could be separated by its own BlocBuilder. 
BUT there are somethings that are not widget. Like `color` and `decoration` in `Container`, or 
the `opacity` value in `Opacity` widget (like the example in #1752 mentioned)

Or assuming in a stack or column which already have 10 widget and 9 of them needs BlocBuilder but one of them doesn't (and it's really heavy!). In case of separating each one to a separated BlocBuilder, it reduces boilerplate code.

felangel:
> > Can you provide a real-world example of when you'd use `child`? I'm not sure I understand when this type of optimization would be beneficial. It makes sense for widgets like `AnimatedBuilder` because you are dealing with animations but I'm not convinced it makes sense in the context of bloc. In addition, you can always create your own widget that wraps `BlocBuilder` to apply this change and I'd prefer not to include this unless there's a very strong argument in its favor. Let me know what you think, thanks!
> 
> In flutter most of things are widgets so they could be separated by its own BlocBuilder. BUT there are somethings that are not widget. Like `color` and `decoration` in `Container`, or the `opacity` value in `Opacity` widget (like the example in #1752 mentioned)
> 
> Or assuming in a stack or column which already have 10 widget and 9 of them needs BlocBuilder but one of them doesn't (and it's really heavy!). In case of separating each one to a separated BlocBuilder, it reduces boilerplate code.

Can you provide some code snippets that illustrate the scenario you have in mind?

mahdi739:
>
> Can you provide some code snippets that illustrate the scenario you have in mind?

This is a part my code I'm working on it right now:

```
Stack(
                alignment: Alignment.bottomCenter,
                children: [
                  Container(
                    constraints: const BoxConstraints.expand(height: double.infinity),
                    // child: FadeTransition(
                    // opacity: controller.animationFirst,
                    child: controller.firstImageProvider.value,
                  ), //),
                  Container(
                      constraints: const BoxConstraints.expand(height: double.infinity),
                      child: FadeTransition(
                        opacity: controller.animationSecond,
                        child: controller.secondImageProvider.value,
                      )),
                  if (controller.filesController.settingsService.themeIndex.val != 1 &&
                      controller.filesController.settingsService.themeIndex.val != 3)
                    AnimatedContainer(
                      duration: Duration(milliseconds: 300),
                      height: 300,
                      alignment: Alignment.bottomCenter,
                      decoration: BoxDecoration(
                          gradient: LinearGradient(
                              begin: Alignment.bottomCenter,
                              end: Alignment.topCenter,
                              colors: [
                            controller.filesController.realDominantColor.value,
                            controller.filesController.realDominantColor.value.withOpacity(0)
                          ])),
                    ),
                  if (controller.filesController.settingsService.themeIndex.val == 1)
                    Container(
                      color: Colors.black.withOpacity(0.6),
                    ),
                  if (controller.filesController.settingsService.themeIndex.val == 3)
                    Container(
                      color: Colors.white.withOpacity(0.6),
                    ),
                  container,

                  
                ],
              ),
```

(Actually it is already written with another state management but I'm migrating to Bloc. 🙂)

As it shows, the stack has 6 children which 5 of them depend on state but the last one `container` does not (which is my content of page). For now I cached it to a pre-built widget, but it prevent me to make constructor of page `const` and it will become messy if I try to make a separated BlocBuilder for each one.

felangel:
@mahdi739 in the example you provided I recommend specifying a const instance of the child like:

```dart
BlocBuilder<FilesBloc, FilesState>(
              builder: (context, state) {
                return Container(
                  margin: const EdgeInsets.all(10),
                  height: MediaQuery.of(context).size.height * 0.8,
                  width: MediaQuery.of(context).size.width * 0.9,
                  decoration: BoxDecoration(
                    borderRadius: BorderRadius.circular(15),
                    border: Border.all(
                        color: calculateColor(state.imageSource), width: 4),
                    image: DecorationImage(
                        image: FileImage(File(state.imageSource))),
                  ),
                  child: Container(
                    decoration: BoxDecoration(
                      borderRadius: BorderRadius.circular(15.0),
                    ),
                    clipBehavior: Clip.antiAlias,
                    child: BackdropFilter(
                      filter: ImageFilter.blur(
                        sigmaX: 20,
                        sigmaY: 20,
                      ),
                      child: const ContentWidget(), //Doesn't depend on state
                    ),
                  ),
                );
              },
            )
```

This will ensure the widget is not rebuilt unnecessarily.

mahdi739:
>
> This will ensure the widget is not rebuilt unnecessarily.

Thank you. Yes but there is some situations that we are not able to use const constructor because that widget doesn't have const constructor or the variables we want to pass it are not constant.
But actually I think if we have this kind of situations, the problem comes from bad coding practices.

mahdi739:
> > > This will ensure the widget is not rebuilt unnecessarily.
> > 
> > 
> > Thank you. Yes but there is some situations that we are not able to use const constructor because that widget doesn't have const constructor or the variables we want to pass it are not constant. But actually I think if we have this kind of situations, the problem comes from bad coding practices.
> 
> Yeah in most cases you should be able to return a const instance. If you have some examples in which thats not possible I’m happy to take a look and discuss alternatives.
> 
> Based on our conversation do you still feel these changes are needed?

This situation occurs mostly when we get some variables from the class and want to pass it to the child widget (that doesn't depend on the state).

for above example it becomes like this:

```
 class ParentWidget extends StatelessWidget {
  const ParentWidget({Key? key, required this.data}) : super(key: key);
  final String data;
  @override
    return BlocBuilder<FilesBloc, FilesState>(
      builder: (context, state) {
        return Container(
          margin: const EdgeInsets.all(10),
          height: MediaQuery.of(context).size.height * 0.8,
          width: MediaQuery.of(context).size.width * 0.9,
          decoration: BoxDecoration(
            borderRadius: BorderRadius.circular(15),
            border: Border.all(color: calculateColor(state.imageSource), width: 4),
            image: DecorationImage(image: FileImage(File(state.imageSource))),
          ),
          child: Container(
            decoration: BoxDecoration(
              borderRadius: BorderRadius.circular(15.0),
            ),
            clipBehavior: Clip.antiAlias,
            child: BackdropFilter(
              filter: ImageFilter.blur(
                sigmaX: 20,
                sigmaY: 20,
              ),
              child: ContentWidget(data: data), //Doesn't depend on the state and needs data!
            ),
          ),
        );
      },
    );
  }
}
```


and in another example we have a CheckboxListTile and its value depends on the state but its secondary doesn't. However we can't use const for that:

```
class MyTile extends StatelessWidget {
  const MyTile({
    Key? key,
    required this.height,
    required this.width,
    required this.zentity,
    required this.selectBloc,
    this.onTap,
  }) : super(key: key);
  final double height;
  final double width;
  final Zentity zentity;
  final SelectBloc selectBloc;
  final Function? onTap;

  @override
    return BlocBuilder<SelectBloc, SelectState>(
      bloc: selectBloc,
      builder: (context, state) {
        return CheckboxListTile(
          onChanged: (bool? value) {
            //...
          },
          value: state.isSelected,
          secondary: ImageThumb(
            zentity,
            height: height,
            width: width,
          ),//Doesn't depend on the state and needs data!
        );
      },
    );
  }
}
```

According to these examples (and the first example in our conversation which was a column with 6 children) I still think it's needed in this kind of situations.

felangel:
@mahdi739 in that case why not just declare the widget outside the `BlocBuilder` like:

```dart
class MyTile extends StatelessWidget {
  const MyTile({
    Key? key,
    required this.height,
    required this.width,
    required this.zentity,
    required this.selectBloc,
    this.onTap,
  }) : super(key: key);
  final double height;
  final double width;
  final Zentity zentity;
  final SelectBloc selectBloc;
  final Function? onTap;

  @override
    final thumbail =  ImageThumb(
        zentity,
        height: height,
        width: width,
      ),
    );
    return BlocBuilder<SelectBloc, SelectState>(
      bloc: selectBloc,
      builder: (context, state) {
        return CheckboxListTile(
          onChanged: (bool? value) {
            //...
          },
          value: state.isSelected,
          secondary: thumbnail,
      },
    );
  }
}
```

mahdi739:
> @mahdi739 in that case why not just declare the widget outside the `BlocBuilder` like:
> 
> ```dart
> class MyTile extends StatelessWidget {
>   const MyTile({
>     Key? key,
>     required this.height,
>     required this.width,
>     required this.zentity,
>     required this.selectBloc,
>     this.onTap,
>   }) : super(key: key);
>   final double height;
>   final double width;
>   final Zentity zentity;
>   final SelectBloc selectBloc;
>   final Function? onTap;
> 
>   @override
>   Widget build(BuildContext context) {
>     final thumbail =  ImageThumb(
>         zentity,
>         height: height,
>         width: width,
>       ),
>     );
>     return BlocBuilder<SelectBloc, SelectState>(
>       bloc: selectBloc,
>       builder: (context, state) {
>         return CheckboxListTile(
>           onChanged: (bool? value) {
>             //...
>           },
>           value: state.isSelected,
>           secondary: thumbnail,
>       },
>     );
>   }
> }
> ```

Yes it actually does the same thing. So I don't think it's necessary to have an extra child parameter. 
Thank you for helping 🙂

PR #3291, opened 2022-03-24
feat(flutter_bloc): provide all Multi*Provider parameters in constructor

tynn:
## Status

**READY**

## Breaking Changes

NO (compiletime) / maybe YES (binary)

## Description
Making the child nullable to simplify usage within a _nested_ hierarchy.
Also adding the transition builder for consitency with `MultiProvider`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/lib/src/multi_bloc_listener.dart
+++ b/packages/flutter_bloc/lib/src/multi_bloc_listener.dart
@@ -52,6 +52,7 @@ class MultiBlocListener extends MultiProvider {
   MultiBlocListener({
     Key? key,
     required List<BlocListenerSingleChildWidget> listeners,
-    required Widget child,
-  }) : super(key: key, providers: listeners, child: child);
+    required Widget? child,
+    TransitionBuilder? builder,
+  }) : super(key: key, providers: listeners, child: child, builder: builder);
 }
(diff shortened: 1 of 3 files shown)


felangel:
Hi @tynn 👋 
Thanks for opening an issue!

Can you please provide a bit more context regarding cases in which having a `TransitionBuilder` is beneficial? I completely understand the argument about maintaining parity with `package:provider` but I also want to make sure I fully understand the use-case we're trying to solve by making this change to the API.

Thanks!

felangel:
> Hej @felangel, parity was actually the only reason. Should I just remove it?
> 
> The issue I had was mostly with `Widget child`, forcing me use `MultiProvider` instead of `MultiBlocProvider` or assume `child!` within a nested hierarchy.

Do you mind providing a short snippet of code to illustrate the issue you're describing? I want to make sure I fully understand, thanks!

tynn:
@felangel could you let me know which parts you'd want to merge? I'll update the PR accordingly or close it then.

felangel:
If the only reason is feature parity I'd highly prefer to keep the API as is because MultiBlocProvider's child should not be optional imo.

PR #3467, opened 2022-08-01
Flutter login formz update

georgiossalon:
Updated the flutter_login example to work with the updated formz ^0.5.0-dev.1 package.

## Status

**READY**

## Breaking Changes

NO

## Description

By using the new formz package you may omit the Formz.validate function (which does not exist in the newest version). I used the FormzMixin on the LoginState class. This way I can directly get from the state whether the login form is valid or not.

## Type of Change

- [x ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


georgiossalon:
@felangel Is there something missing that I should change in this PR?

examples/flutter_login/lib/login/bloc/login_bloc.dart
```diff
@@ -27,7 +27,6 @@ class LoginBloc extends Bloc<LoginEvent, LoginState> {
     emit(
       state.copyWith(
         username: username,
-        status: Formz.validate([state.password, username]),
```
LostInDarkMath, line 30: Can you explain please why this is no longer necessary?

myxzlpltk: In new version, formz package did breaking changes by **decouple purity, validity, and submission status**. There also rename `FormzStatus` to `FormzSubmissionStatus`. Property status inside LoginState from now on will only reflect on submission status. On the other hand, validity and purity will be accessible via getter inside LoginState after implement `FormzMixin`.

Since `Formz.validate()` will return a boolean while status is FormzSubmissionStatus and accessible via getter. We don't need to refresh status property. We can access it on demand via `state.isValid`.

examples/flutter_login/lib/login/view/login_form.dart
```diff
@@ -80,13 +80,13 @@ class _LoginButton extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return BlocBuilder<LoginBloc, LoginState>(
-      buildWhen: (previous, current) => previous.status != current.status,
+      buildWhen: (_, current) => current.isValid,
```
myxzlpltk, line 83: I don't think this is a correct way to define it. Login button won't rebuild if login form not valid, therefore button cannot disable after enabled for example by erase all input. And it's rebuild when isValid, even though type single character will rebuild the button.

Here is my suggestion:
```dart
buildWhen: (previous, current) =>
  previous.status != current.status ||
  previous.isPure != current.isPure ||
  previous.isValid != current.isValid,
```

myxzlpltk, line 83: I don't think this is a correct way to define it. Login button won't rebuild if login form not valid, therefore button cannot disable after enabled for example by erase all input. And it's rebuild when isValid, even though type single character will rebuild the button.

Here is my suggestion:
```dart
buildWhen: (previous, current) =>
  previous.status != current.status ||
  previous.isPure != current.isPure ||
  previous.isValid != current.isValid,
```

examples/flutter_login/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
     sdk: flutter
   flutter_bloc:
     path: ../../packages/flutter_bloc
-  formz: ^0.4.0
+  formz: ^0.5.0-dev.1
```
myxzlpltk, line 17: You can upgrade it to 0.6.0
`formz: ^0.6.0`

examples/flutter_login/lib/login/bloc/login_state.dart
```diff
@@ -25,4 +25,9 @@ class LoginState extends Equatable {
 
   @override
   List<Object> get props => [status, username, password];
```
myxzlpltk, line 27: After decoupling purity, validity, and submission status. This will be the appropriate way to define equatable's props
```dart
List<Object> get props => [status, username, password, isValid, isPure];
```

myxzlpltk (changes requested):
Almost 8 months huh? Here is a new suggestions from me

examples/flutter_login/lib/login/view/login_form.dart
```diff
@@ -10,7 +10,7 @@ class LoginForm extends StatelessWidget {
   Widget build(BuildContext context) {
     return BlocListener<LoginBloc, LoginState>(
       listener: (context, state) {
```
v1ncere, line 12: add listenWhen because after an exception when you correct your input it'll always call the snackbar.
`listenWhen: (previous, current) => previous.status != current.status,`

felangel:
This has been done as part of the Dart 3 migration #3809

PR #3492, opened 2022-08-11
Added onDone in forEach and onEach methods of Emitter

pedrolemoz:
## Status

**READY**

## Breaking Changes

**NO**

## Description

In the current version of `Bloc`, if I want to emit a "finishing state" when a Stream is closed, I would do the following:

image

It would be much easier to figure this out if we have the same callbacks of the [`listen`](https://api.flutter.dev/flutter/dart-async/Stream/listen.html) method of a `Stream`, which includes the `onDone` callback.

So, I added this callback on `forEach` and `onEach` methods of `Emitter`, and now, we can do this instead:

image

The first approach doesn't ensure that the `Stream` is really closed, but by relying on `onDone` callback from `listen`, we can guarantee this. We can verify this behavior in the unit test, where the `close` method of `StreamController` is called, and `onDone` is invoked. Is just a syntactic sugar, but it can be really useful when working with `Stream` through `Bloc`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/emitter.dart
+++ b/packages/bloc/lib/src/emitter.dart
@@ -23,11 +23,14 @@ abstract class Emitter<State> {
   /// [onError] and will not result in unhandled exceptions or cancelations to
   /// the internal stream subscription.
   ///
+  /// If [onDone] is provided, when the [stream] ends, [onDone] will be invoked.
+  ///
   /// **Note**: The stack trace argument may be [StackTrace.empty]
   /// if the [stream] received an error without a stack trace.
   Future<void> onEach<T>(
     Stream<T> stream, {
     required void Function(T data) onData,
+    void Function()? onDone,
     void Function(Object error, StackTrace stackTrace)? onError,
   });
 
@@ -45,11 +48,14 @@ abstract class Emitter<State> {
   /// [onError] and will not result in unhandled exceptions or cancelations to
   /// the internal stream subscription.
   ///
+  /// If [onDone] is provided, when the [stream] ends, [onDone] will be invoked.
+  ///
   /// **Note**: The stack trace argument may be [StackTrace.empty]
   /// if the [stream] received an error without a stack trace.
   Future<void> forEach<T>(
     Stream<T> stream, {
     required State Function(T data) onData,
+    State Function()? onDone,
     State Function(Object error, StackTrace stackTrace)? onError,
   });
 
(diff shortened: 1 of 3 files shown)


felangel:
Hi @pedrolemoz 👋 
Thanks for the pull request!

While I understand wanting to maintain the callbacks as `listen`, I prefer not to introduce having multiple ways of achieving the same thing because it introduces unnecessary complexity.

> The first approach doesn't ensure that the Stream is really closed, but by relying on onDone callback from listen, we can guarantee this

Can you elaborate on this? Based on the current implementation, the `onDone` callback is invoked at the exact same time as when the completer is completed so I don't see how there would be any difference in behavior.

pedrolemoz:
Thanks for the response @felangel.

I didn't knew about the usage of `Completer`, so I took a time to learn about it. You're right, there's no difference in behavior. Also, there's nothing stopping the user to emit a new `state` in both ways, but `emit` method will perform an early return if the `state` is duplicated. No `state` change should occur.

https://github.com/felangel/bloc/blob/6b61110e5b72c6183e85c17f507ce205383592df/packages/bloc/lib/src/bloc_base.dart#L98

I understand your point, but I still prefer using a callback inside `forEach/onEach` instead of using `emit` in another line. In my opinion, it improves readability and makes easy for the user to emit a "finishing state", since everything is happening in one place. `Bloc` is known for being complex and very restrictive in some aspects (you should not have `BuildContext` inside your `Bloc`, or even call a function directly, but use `.add` instead, just to give some examples). Some people have fear of `Streams`, but they are a little bit comfortable with the Dart Async API. After all, it's just a simple improvement, and not a breaking change. I think this change makes sense, even if don't make a huge impact.

pstromberg98:
I agree with having a single way to do things.

Due to the `onEach` having an `onError` and `onData` I would expect an `onDone` due to what almost seems to be an interface parity with steams. To optimize for what the user *expects*, I think you would want either:

1. Remove `onError` and have the returned `Future` throw. **This would make `cancelOnError: false` cases impossible**.
or
2. Achieve full interface parity with `.listen` by adding `onDone`, optionally make `onData` a positional argument, and remove the returned `Future` all together. It may also make sense to rename `forEach` to `listen` or `stream`.

The first case prevents a use case so that isn't viable. 
The second one removes the future which has a minor con in my opinion as having a returned `Future` is more robust and more easily supports other use cases.

All of that being said, both of those options introduce breaking changes and probably aren't worth it.

felangel:
Closing this because I don't think this change is worth it. As a I mentioned above it introduces multiple ways to achieve the same thing and it also would be unclear whether onDone would be called in all cases (completion + error or just completion).

Issue #3377, opened 2022-05-28
refactor: bloc_test to depend on core interfaces rather than BlocBase realization


tsinis:
**Description**

Hey! Thanks for a great package, I have a sort of feature request for refactoring :D

At this moment, `blocTest` from `bloc_test` package is [dependent on BlocBase](https://github.com/felangel/bloc/blob/master/packages/bloc_test/lib/src/bloc_test.dart#L139) realization, but since #3012 we have interfaces I believe it should be dependent on, those interfaces, following the SOLID, Dependency Inversion Principle (DIP). For example, it will bring an ability to use the `bloc_test` package for testing custom blocs that are implementing those interfaces, not the `BlocBase`. Of course, it's upon you. Thanks again!

purplenoodlesoop:
As someone who took part in the original "core interfaces" discussion – yes, please!

felangel:
I think this is a great idea as well but it'd be nice to discuss what that would look like. Currently, it's not enough to just change `BlocBase` to `StateStreamable` because `blocTest` uses `emit` and `close` internally. This means the interface would have to implement `StateStreamable`, `Emittable`, and `Closable`.

I'd love to hear what you had in mind, thanks!

tsinis:
> 

Hey @felangel, thanks for the quick response. I must confess that at the moment I have no concrete idea how to implement it properly without breaking changes, otherwise, it would not be a refactor suggestion but a regular Pull Request :). I just noticed that the blockTest doesn't use interfaces and was surprised. Yes, most likely it would have to be done through references on both `StateStreamableSource` and `Emittable`. Unfortunately, I won't have time in the next weeks as I'm starting at the new job on Wednesday. So, are you free to close this issue now, and I can come back to this point later and open a new one with a more specific implementation or with a PR? In any case, thanks for your attention.

felangel:
@tsinis sorry for the confusion. I don’t think we should close this issue but I just think we need to discuss the naming, structure of the interfaces before opening any pull requests, thanks!

tsinis:
> feat(bloc): introduced BlocBaseInterface #3543

Hi @felangel, I understand, but unfortunately, I haven't found a proper reference to the names of the BLoC interfaces anywhere and I am definitely the worst possible person to name it. Do you have any ideas about the naming and the general abstraction idea?

felangel:
> > feat(bloc): introduced BlocBaseInterface #3543
> 
[...]

PR #3543, opened 2022-09-16
feat(bloc): introduced BlocBaseInterface

tsinis:
## Status

**READY**

## Breaking Changes

NO

## Description

Hey @felangel! The pull request is directly related to #3377, and because abstractions are no longer discussed there, I've created this absolutely straightforward solution. This new `BlocBaseInterface` can be later used in the `bloc_test` package, to make it more abstract:

```dart
void blocTest<B extends BlocBaseInterface<State>, State>

...

Future<void> testBloc<B extends BlocBaseInterface<State>, State>
```

Then all inheritors of pure bloc (that need to have their blocs testable using the `bloc_test` package, for example) can rely on this interface and not on `BlocBase` itself. Because in most cases they already inherit from both `Emittable<State>` and `StateStreamableSource<State>`. It's a rather primitive solution, but in my opinion better than nothing. I'm open to discussion, but also feel free to close this PR without comment, if you don't like this approach. Thanks!

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/bloc_base.dart
+++ b/packages/bloc/lib/src/bloc_base.dart
@@ -44,12 +44,16 @@ abstract class ErrorSink implements Closable {
   void addError(Object error, [StackTrace? stackTrace]);
 }
 
+/// The basic interface with minimal access to both:
+/// synchronous access to the current [state] and states emitting.
+abstract class BlocBaseInterface<State>
+    implements StateStreamableSource<State>, Emittable<State> {}
+
 /// {@template bloc_base}
 /// An interface for the core functionality implemented by
 /// both [Bloc] and [Cubit].
 /// {@endtemplate}
-abstract class BlocBase<State>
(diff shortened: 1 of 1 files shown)


packages/bloc/lib/src/bloc_base.dart
```diff
@@ -44,12 +44,16 @@ abstract class ErrorSink implements Closable {
   void addError(Object error, [StackTrace? stackTrace]);
 }
 
+/// The basic interface with minimal access to both:
+/// synchronous access to the current [state] and states emitting.
+abstract class BlocBaseInterface<State>
```
felangel, line 49: I think this requires further discussion before making any code changes. I'd prefer not to add `Interface` as a suffix for example since that is inconsistent with other interfaces within the library. Closing this PR for now until we can come up with a naming convention that is generally agreed upon and consistent with the current codebase.

Issue #3418, opened 2022-07-03
feat: BlocSelector with "selectWhen"


dedonnodev:
**Description**

There is no option to select which Bloc's event should trigger the "selector" of BlocSelector

**Desired Solution**

it would be awesome to have a "selectWhen" in BlocSelector, to avoid useless rebuild with empty values

**Additional Context**

Now i use check in "selector" if the event is the event i'm interested of, and if it isn't, i return a null/empty value that trigger a rebuild of child (that i wouldn't need)

felangel:
Hi @dedonnodev 👋
Thanks for opening an issue!

The UI should only depend on the state. The presentation layers adds events to the bloc and can listen for state changes. If you are rebuilding the UI depending on the event added then that’s no longer the bloc pattern.

 I recommend refactoring so that all the information needed is provided via the bloc state. Can you provide a concrete use-case for why your presentation layer rebuilds depending on the event? Thanks!

felangel:
> Hi @felangel I'm using Bloc for a "settings page", where every section rebuild when its setting value change. I'm using a BlocSelector. Example: `BlocSelector<SettingsBloc, SettingsState, String>( selector: (state) => state.maybeMap( ready: (value) => value.printerIp, orElse: () => '', ), builder:{...} )`
> 
> i use a SettingBloc for the whole "settings page" and then use blocSelector for "filter" settings for each section. But i don't think it's the right way... isn't it? 🥲

Can you provide a link to a minimal reproduction sample that illustrates the issue? It would be much easier to help if I can reproduce locally, thanks! 🙏

felangel:
Closing for now since there is no reproduction sample.

bekaIva:
Can we reopen this please?

This is my issue with BlocSelector.

The Bloc:
`class LiveStreamBloc extends Bloc<LiveStreamEvent, LiveStreamState> {
  final GetLiveStream _getLiveStream;
  LiveStreamBloc(this._getLiveStream) : super(const LiveStreamState.initial()) {
    on<LiveStreamEvent>((event, emit) async {
      await switch (event) {
        StartFetchingLiveStream() => emit.forEach(
            _mapStartFetchingLiveStreamToState(emit),
            onData: (data) => data,
            onError: (error, stackTrace) {
              return LiveStreamState.error(error.toString());
            },
          ),
      };
    }, transformer: restartable());
  }

[...]

PR #3602, opened 2022-11-03
add `BlocSelectorCondition` to `BlocSelector`

Wouter12:
## Status
Ready

## Breaking Changes

<s>YES</s> | NO

## Description

This is a PR that allows for more control over how often the BlocSelector rebuilds, this solves #3418

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/README.md
+++ b/packages/flutter_bloc/README.md
@@ -180,6 +180,22 @@ BlocSelector<BlocA, BlocAState, SelectedState>(
 )
 ```
 
+For fine-grained control over when the `builder` function is called an optional `selectWhen` can be provided. `selectWhen` takes the previous bloc state and current bloc state and returns a boolean. If `selectWhen` returns true, `builder` will be called with `state` and the widget will rebuild. If `selectWhen` returns false, `builder` will not be called with `state` and no rebuild will occur.
+
+```dart
+BlocSelector<BlocA, BlocAState, SelectedState>(
+  selector: (state) {
+    // return selected state based on the provided state.
+  },
+  selectWhen: (previousState, state) {
+    // return true/false to determine whether or not
+    // to rebuild the widget with state
+  },
+  builder: (context, state) {
+    // return widget here based on selected state
+  }
+)
+
 ### BlocProvider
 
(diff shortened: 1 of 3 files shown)


felangel:
Hi @Wouter12 👋 
Thanks for opening a PR however I'm not convinced this change makes sense. Would you be able to provide additional context for when this would be useful via a sample application? I feel this API would be quite confusing because we would then have two mechanisms to regular rebuilds within a single widget.

Let me know what you think and thanks again!

Wouter12:
Imagine a widget which shows the user's last known location. This location is fetched with a `LocationBloc`.
This bloc needs to handle the location service being unavailable, and permissions being asked. 

This is handled in a completely different widget, and the widget only shows the last known location. 

instead of handling all the states in the selector, and work with a nullable selectorstate, you can use selectWhen to only allow the LocationKnown state pass
to the selector, and cast it there, which result in this
``` dart
BlocSelector<LocationBloc, LocationState, Location>(
  selector: (state) => (state as LocationKnown).location,
  selectWhen: (_, current) => current is LocationKnown,
  builder: (context, state) => Text('location is ${state.toString()'),
);
``` 
instead of this
``` dart
BlocSelector<LocationBloc, LocationState, Location?>(
  selector: (state) {
    if(state is LocationKnown) {
      return state.location;
    }
    return null;
  },
  builder: (context, state) {
    if(state != null) {
      return Text('location is ${state.toString()');
    } else {
      return Text('unknown location');
  },
);
```

felangel:
> ```dart
>   selectWhen: (_, current) => current is LocationKnown,
> ```

The issue is the conditions like `buildWhen` are purely an optimization and they do not guarantee that the `builder` will not be called. For example, even if your `buildWhen` always returned `false`, the `builder` would still need to be called at least once. In addition, if the parent widget rebuilds, the `builder` will also run even if the condition evaluates to false. For these reasons I prefer not to introduce this change because `BlocSelector` already reduces rebuilds.

Closing for now but feel free to comment with any additional feedback/thoughts and I'm happy to continue the conversation 👍

PR #3751, opened 2023-02-23
docs(architecture): improve `blocs_presentation.dart.md` snippet

lsaudon:
## Status

**READY**

## Breaking Changes

NO

## Description

`FirstBloc, FirstState` is more consistent in the example compared to  `WeatherCubit, WeatherState`

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/docs/_snippets/architecture/blocs_presentation.dart.md
+++ b/docs/_snippets/architecture/blocs_presentation.dart.md
@@ -4,7 +4,7 @@ class MyWidget extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return BlocListener<WeatherCubit, WeatherState>(
+    return BlocListener<FirstBloc, FirstState>(
       listener: (context, state) {
         // When the first bloc's state changes, this will be called.
         //
--- a/docs/_snippets/architecture/blocs_presentation.dart.md
+++ b/docs/_snippets/architecture/blocs_presentation.dart.md
@@ -4,18 +4,18 @@ class MyWidget extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return BlocListener<WeatherBloc, WeatherState>(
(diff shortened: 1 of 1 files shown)


tenhobi:
Hi, how is that more consistent? `WeatherCubit` sound more like what it should do, so it is the right name IMO. What should maybe change are the names of other examples etc., if something more expressing could be used. But I see no issue in this case.

lsaudon:
This was just because WeatherCubit does not exist in the example.
Indeed, maybe do the opposite, find a more concrete name for SecondBloc.

lsaudon:
Maybe better?
- `ClothingBloc`, `ClothingBlocEvent`
- `WeatherBloc`, `RefreshWeather`

docs/_snippets/architecture/blocs_presentation.dart.md
```diff
@@ -4,18 +4,18 @@ class MyWidget extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return BlocListener<WeatherCubit, WeatherState>(
+    return BlocListener<WeatherBloc, WeatherState>(
```
felangel, line 7: If we make this change we will also need to update the corresponding documentation in https://github.com/felangel/bloc/blob/b11fe6a1c00a5cb22fd38b2eba2ae6c0c773cec5/docs/architecture.md?plain=1#L73 (and other translations). The simpler change would be to go back to using `FirstBloc` and `SecondBloc` in order to be consistent with the rest of the docs. Let me know what you think and thanks for the contribution! 🙏

felangel (changes requested):
See my comment and lmk what you think.

tenhobi (approved):

felangel (commented):
LGTM thanks so much for the contribution! 💯

Issue #3756, opened 2023-03-07
Bad state: Cannot emit new states after calling close


Lazizbek97:
**Description**
I've been using Cubit&Freezed for the app state management. I have wrapped MaterailApp with MultiBlocProvider and initialized all providers there. When the user logout, the app should remove all pages and push them to the login page.  Here is the Problem. My main cubit is being disposed when i navigate with pushAndRemoveUntill() to login page, and after user login again, the app crashes with the error: 
``` Bad state: Cannot emit new states after calling close
flutter: 
#0      BlocBase.emit (package:bloc/src/bloc_base.dart:96:9)
#1      MainCubit._updateCounterEnabled (package:billz_new_app/src/modules/main/cubit/main_cubit.dart:128:5)
#2      MainCubit.unleashInit.<anonymous closure> (package:billz_new_app/src/modules/main/cubit/main_cubit.dart:119:7)
#3      EventListener.call (package:events_emitter/listener.dart:61:33)
#4      EventEmitter.emitEvent (package:events_emitter/emitters/event_emitter.dart:93:34)
```
**Expected Behavior**
When the user logout and login, I want to refresh all cubits according to the user's credentials. without any crashes.

tenhobi:
Hey @Lazizbek97, could you send an example of how you setup your bloc provider and the `MaterialApp`?

sanya-wadhwa:
I am not able to run void main for test cases as it is showing **Bad state: Cannot add new events after calling close**..
Any update or solution for this?
Here is the full error:

_dart:async                                   _BroadcastStreamController.add
package:bloc/src/bloc.dart 99:24             Bloc.add
test/video_news/video_news_test.dart 128:16  main.<fn>.<fn>
package:bloc_test/src/bloc_test.dart 208:20  testBloc.<fn>
===== asynchronous gap ===========================
dart:async                                   _CustomZone.registerUnaryCallback
package:bloc_test/src/bloc_test.dart 201:7   testBloc.<fn>
dart:async                                   runZonedGuarded
package:bloc_test/src/bloc_test.dart 199:9   testBloc
package:bloc_test/src/bloc_test.dart 156:13  blocTest.<fn>

Bad state: Cannot add new events after calling close_

tenhobi:
Hey @sanya-wadhwa, could you send us a minimal reproducible example with the error?

MateusHBR:
[...]

PR #3789, opened 2023-04-04
FIX: unprocessed events added to eventStream when isClosed is false

MateusHBR:
## Status
READY

## Breaking Changes
NO

## Description

https://github.com/felangel/bloc/issues/3756

This PR fixes the issue in the comment, the problem is that the Bloc is closed during the event processing.

I'm not sure if this is the best solution, but maybe set the isClosed when the close be called could be better. The solution solves the problem because the event is just closed after the emitters, and is processed.

> 
> I'm also encountering the same issue. In my implementation, I have a widget tree with a Screen widget that contains a MultiBlocProvider with a CustomTabBarView as its child. Each CustomTabBarView has some blocs that are necessary for it to function properly. In the dispose method of the widget, I add a new event to the blocs to perform some necessary cleanup.
> 
> However, when attempting to add an event to another bloc inside one of the blocs, a StateError is thrown because the event stream appears to be closed. This is problematic because the isClosed property of the bloc always returns false, as it only checks the StateStream and not the eventStream.
> 
> My Widget Tree:
> ```
> Screen(
>    TabBuilder(
>      MultiBlocProvider(
>        providers: [
>            BlocProvider<BlocA>(),
>            BlocProvider<BlocB>(),
>            ...
>        ],
>        child: CustomTabBarView()
>      )
>    )
> )
> ```

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following StateError was thrown running a test:
Bad state: Cannot add new events after calling close

When the exception was thrown, this was the stack:
#1      Bloc.add (package:bloc/src/bloc.dart:99:24)
#2      FeedItemBloc._mapFeedItemPausedToState (package:indaband/feed/widgets/feed_item_view/bloc/feed_item_bloc.dart:639:20)
#3      Bloc.on.<anonymous closure>.handleEvent (package:bloc/src/bloc.dart:226:26)
#4      Bloc.on.<anonymous closure> (package:bloc/src/bloc.dart:235:9)
#13     CastStreamSubscription._onData (dart:_internal/async_cast.dart:85:11)
#41     FakeAsync.flushMicrotasks (package:fake_async/fake_async.dart:197:32)
#42     AutomatedTestWidgetsFlutterBinding.scheduleWarmUpFrame (package:flutter_test/src/binding.dart:1283:24)
#43     runApp (package:flutter/src/widgets/binding.dart:1019:7)
#44     TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:942:7)
<asynchronous suspension>
(elided 36 frames from dart:async and package:stack_trace)
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -277,12 +277,12 @@ abstract class Bloc<Event, State> extends BlocBase<State>
   @mustCallSuper
   @override
   Future<void> close() async {
-    await _eventController.close();
     for (final emitter in _emitters) {
       emitter.cancel();
     }
     await Future.wait<void>(_emitters.map((e) => e.future));
     await Future.wait<void>(_subscriptions.map((s) => s.cancel()));
+    await _eventController.close();
     return super.close();
   }
 }
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -128,6 +128,9 @@ abstract class Bloc<Event, State> extends BlocBase<State>
     _blocObserver.onEvent(this, event);
   }
 
+  @override
+  bool get isClosed => super.isClosed || _eventController.isClosed;
+
   /// {@template emit}
   /// **[emit] is only for internal use and should never be called directly
   /// outside of tests. The [Emitter] instance provided to each [EventHandler]
@@ -277,12 +280,12 @@ abstract class Bloc<Event, State> extends BlocBase<State>
   @mustCallSuper
   @override
   Future<void> close() async {
+    await _eventController.close();
     for (final emitter in _emitters) {
       emitter.cancel();
     }
     await Future.wait<void>(_emitters.map((e) => e.future));
     await Future.wait<void>(_subscriptions.map((s) => s.cancel()));
-    await _eventController.close();
     return super.close();
   }
 }


tenhobi (commented):
I disagree. When a bloc is marked as to be closed, and the `close` method is called, it should not produce any new events. Therefore the `_eventController.close()` should be the first thing.

The mentioned issues are caused by improper use rather than by a bug in Bloc library.

MateusHBR:
@tenhobi WDTY about the first thing that should be done is mark the Bloc closed when the close be called? With this change this could be safely handled, like checking if the Bloc is closed to add or not a event.

With this change, it will works similar with the mounted from a stateful widget that is marked as unmounted when the element will be removed from the tree.

tenhobi:
Isn't `_eventController.close()` marking a bloc as closed? The rest of the code in the `close` function, as the docs say, only manages pending events and states.

>   /// Closes the `event` and `state` `Streams`.
  /// This method should be called when a [Bloc] is no longer needed.
  /// Once [close] is called, `events` that are [add]ed will not be
  /// processed.
  /// In addition, if [close] is called while `events` are still being
  /// processed, the [Bloc] will finish processing the pending `events`.

tenhobi:
The event stream should be closed right away; that's fine by me. 

Yet I am not sure about the `isClosed` is marked properly. 🤔

Right now, the event stream is closed, then the pending states/events are processed, and then the pending events are processed, and then `super.close` (BlocBase.close) is called when the state controller is closed. 🤔

For sure, you should check if the bloc is closed in the event handler function after any await etc. Yet I think if you can face some issues with correctly used code. "Probably not", since Bloc is used in a lot of big apps with no issue. But maybe we can discuss the order, or just make it clear this is a good order? @felangel

felangel:
Closing this for now since the original issue has been closed and there is no clear minimal reproduction sample to illustrate the issue being fixed here. If you believe this is still a bug please open a new issue with a minimal reproduction sample, thanks! 🙏

PR #3854, opened 2023-05-27
fix(bloc_test): test timeouts due to uncaught exceptions

### LINKED ISSUE #3848: fix: blocTest hangs on unhandled exceptions from flutter 3.10
**Description**
When an exception is thrown in a bloc event then `blocTest` hangs forever (or until the defined timeout). This only happens from flutter 3.10 (dart 3). On 3.7.12 the test fails immediately, which is the expected behavior.

**Steps To Reproduce**

`lib/example_bloc.dart`:

```dart
import 'package:flutter_bloc/flutter_bloc.dart';

class ExampleEvent {
  const ExampleEvent();
}

class ExampleBloc extends Bloc<ExampleEvent, int> {
  ExampleBloc() : super(0) {
    on<ExampleEvent>((event, emit) => throw Exception('intentional error'));
  }
}
```

`test/bloc_test.dart`:

```dart
import 'package:mypackage/example_bloc.dart';

void main() {
  blocTest(
    'unhandled exception',
    build: () => ExampleBloc(),
    expect: () => [0],
    act: (bloc) => bloc.add(const ExampleEvent()),
  );
}
```

**Expected Behavior**
The test fails immediately.

**Additional Context**

I think the problem is that the internal `testBloc` function wraps everything in `runZonedGuarded`. The [docs](https://dart.dev/articles/archive/zones) say errors are not propagated from a zone, which seems to be the case. So when the `test.TestFailure` is thrown from the zone's error handler, it does not go out the zone, so the outer `await` just hangs (if I wrap it in try/catch, that doesn't do anything).

In flutter 3.7.12 the same code works, test fails immediately. So maybe this was changed or bugged in dart 3, not sure.

felangel:
## Status

**HOLD**

## Breaking Changes

NO

## Description

- fix(bloc_test): test timeouts due to uncaught exceptions (closes #3848)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -253,8 +253,8 @@ Future<void> _runZonedGuarded(Future<void> Function() body) {
   runZonedGuarded(() async {
     await body();
     if (!completer.isCompleted) completer.complete();
-  }, (error, stack) {
-    if (!completer.isCompleted) completer.completeError(error);
+  }, (error, stackTrace) {
+    if (!completer.isCompleted) completer.completeError(error, stackTrace);
   });
   return completer.future;
 }
--- a/packages/bloc_test/lib/src/bloc_test.dart
+++ b/packages/bloc_test/lib/src/bloc_test.dart
@@ -196,8 +196,8 @@ Future<void> testBloc<B extends BlocBase<State>, State>({
   );
   Bloc.observer = testObserver;
 
-  await runZonedGuarded(
-    () async {
+  try {
+    await _runZonedGuarded(() async {
       await setUp?.call();
       final states = <State>[];
       final bloc = build();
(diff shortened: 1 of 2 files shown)


packages/bloc_test/lib/src/bloc_test.dart
```diff
@@ -229,26 +229,36 @@ Future<void> testBloc<B extends BlocBase<State>, State>({
...
   if (errors != null) test.expect(unhandledErrors, test.wrapMatcher(errors()));
 }
 
+Future<void> _runZonedGuarded(Future<void> Function() body) {
+  final completer = Completer<void>();
+  runZonedGuarded(() async {
+    await body();
+    if (!completer.isCompleted) completer.complete();
+  }, (error, stack) {
+    if (!completer.isCompleted) completer.completeError(error);
```
slaci, line 257: `completeError()` can take the `stack` too as the 2nd param. I think it would be worthwile to pass if its already there.

felangel: Yeah good call this pr isn’t done yet, still need to add tests and I’ll update to include stackTrace as well 👍

PR #3866, opened 2023-06-06
Improve the UI that matches the flutter_timer tutorial examples

fengwewn:
## Status

READY

## Description

I found that the background in flutter_timer doesn't match the one shown in the tutorial.
Only the linear change of blue and white background.

in  /timer/view/timer.dart
1.in line 131-161,I found the flutter wave code from felangel's previous single page code and modified it to the now background class
web: https://medium.com/flutter-community/flutter-timer-with-flutter-bloc-a464e8332ceb

2.  in line 61-64, And the font color and size of the time have been modified to better match the example 

3.  in line 37-41, And modified the button size to better match the example

in app.dart

4.And modified the button color  to better match the example

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x ] 🗑️ Chore

### DIFF
--- a/examples/flutter_timer/pubspec.yaml
+++ b/examples/flutter_timer/pubspec.yaml
@@ -12,6 +12,7 @@ dependencies:
   flutter:
     sdk: flutter
   flutter_bloc: ^8.1.1
+  wave: ^0.2.2
 
 dev_dependencies:
   bloc_test: ^9.0.0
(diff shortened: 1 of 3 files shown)


examples/flutter_timer/pubspec.yaml
```diff
@@ -12,6 +12,7 @@ dependencies:
   flutter:
     sdk: flutter
   flutter_bloc: ^8.1.1
+  wave: ^0.2.2
```
felangel, line 15: I intentionally removed this package because it was not maintained and appeared to cause lots of performance issues. I’d prefer to update the gif in the docs to match the example personally wdyt? Thanks again for the contribution!

fengwewn: OK, thanks for the reply

PR #3910, opened 2023-07-22
feat(vscode_extension): Dynamic bloc import path base on package in pubspec

haoz33:
## Status

**READY**

## Breaking Changes

NO

## Description

Change `import 'package:bloc/bloc.dart';` to match the package currently used pubspec. For example when `flutter_bloc` is in pubspec, it will use `import 'package:flutter_bloc/flutter_bloc.dart';` instead. This change solves the [depend_on_referenced_packages](https://dart.dev/tools/linter-rules/depend_on_referenced_packages) linting rule when using the extesion with `flutter_bloc` package.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/extensions/vscode/src/commands/new-bloc.command.ts
+++ b/extensions/vscode/src/commands/new-bloc.command.ts
@@ -16,6 +16,7 @@ import {
   getBlocTemplate,
 } from "../templates";
 import { getBlocType, BlocType, TemplateType } from "../utils";
+import { getBlocImportPath } from "../utils/get-bloc-import-path";
 
 export const newBloc = async (uri: Uri) => {
   const blocName = await promptForBlocName();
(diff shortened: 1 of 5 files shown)


felangel:
Hi @haoz33 👋 
Thanks for opening a PR! I'm not sure if we want to encourage importing `flutter_bloc` in a bloc file because the "right" thing to do would be to make sure bloc is added as a dependency to the project. Perhaps it would be better to show a warning if bloc isn't a direct dependency and offer a fix to add it to the pubspec? wdyt?

haoz33:
Hi @felangel 
Sorry for the late reply, was busy on other things last month. I do agree with your point, that is something I didn't think before.

Issue #3555, opened 2022-09-27
feat: Add emitForEach / emitOnEach Methods (or something similar) to Cubit


jtdLab:
**Description**

When working with a `Cubit` that consume some other `Stream` you need to maintain a `StreamSubscription` during the life of the `Cubit` and close it etc. This comes with some extra work which is handled well in `Bloc` via `emit.forEach` / `emit.onEach`  but no similar solution exists for that when using `Cubit`.

**Desired Solution**

Adding the methods `emitForEach` and `emitOnEach` with same signatures and functionality as their counter part in `Emitter`.

**Alternatives Considered**

Maybe give a `Cubit` its own `Emitter`.

jyardin:
I think the alternative solution you proposed of exposing an `Emitter` would be better, as it would give a more consistent experience between `Bloc` and `Cubit`.  

It would be even cooler if the `Emitter` is returned by an `emit` property, replacing the `emit(State)` method.  
It would be a breaking change, not sure how big a breaking change though, as the main usage of `emit` (just calling `emit(newState)`) would still work as intended.

zamaniafshar:
+1

zamaniafshar:
It would be great if Bloc supported this. Until then, the code below might be useful; I just made some changes to the Emitter class.

```dart

base class StreamableCubit<State> extends Cubit<State> {
  StreamableCubit(super.initialState);

  final _completer = Completer<void>();
  final _disposables = <FutureOr<void> Function()>[];

  Future<void> get future => _completer.future;

  @override
  Future<void> close() async {
    for (final disposable in _disposables) {
      disposable.call();
    }
    _disposables.clear();
    if (!_completer.isCompleted) _completer.complete();
    super.close();
  }

  Future<void> onEach<T>(
    Stream<T> stream, {
    required void Function(T data) onData,
    void Function(Object error, StackTrace stackTrace)? onError,
  }) {
    final completer = Completer<void>();
    final subscription = stream.listen(
      onData,
      onDone: completer.complete,
      onError: onError ?? completer.completeError,
      cancelOnError: onError == null,
    );
    _disposables.add(subscription.cancel);
    return Future.any([future, completer.future]).whenComplete(() {
      subscription.cancel();
      _disposables.remove(subscription.cancel);
    });
  }

  Future<void> forEach<T>(
    Stream<T> stream, {
    required State Function(T data) onData,
    State Function(Object error, StackTrace stackTrace)? onError,
  }) {
    return onEach<T>(
[...]

PR #3933, opened 2023-08-15
feat: #3555

jtdLab:
Make `emit` of `BlocBase` return an `Emitter` which is used as an "internal" `Emitter` in `Bloc` replacing the old `emit` function. Inside Cubit the `Emitter` is used instead of the old `emit` function allowing easy handling of subscriptions as described in #3555.

## Status

**READY**

## Breaking Changes

These changes might be breaking for clients which override `emit` (happened in `replay_bloc`).
Else usage should be the same.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel (changes requested):
Unfortunately this is not sufficient because there are APIs and implementation details of the current Emitter implementation that are bloc and event handler specific. Making this work will likely involve splitting Emitter into a base Emitter and a BlocEmitter. I'm looking into this now and will hopefully have something ready to ship in the coming weeks 👍

felangel:
Closing this for now based on my comment but happy to continue the discussion 👍 
Thanks so much for taking the time to contribution -- I really appreciate it! I'll try to get this sorted out very soon!

PR #4080, opened 2024-03-04
docs: fix chinese lang translation

TENX-S:
“中国人” means Chinese people rather than a language.


felangel (approved):
Thanks so much for the contribution! 💙

PR #4144, opened 2024-04-09
docs: Add support kyrgyz language

warioddly:
## Status

**READY**

## Breaking Changes

NO

## Description

Add support kyrgyz language

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -57,6 +57,10 @@ const locales = {
 		label: 'العربية',
 		lang: 'ar',
 	},
+	kg: {
+		label: 'Кыргызча',
+		lang: 'kg',
+	},
 };
 
 // https://astro.build/config


docs/astro.config.mjs
```diff
@@ -57,6 +57,10 @@ const locales = {
 		label: 'العربية',
 		lang: 'ar',
 	},
+	kg: {
```
felangel, line 60: It would be nice to include some translations as part of this change even if it's just for the index.mdx. Otherwise this change won't provide as much value to users since it'll just add the item to the translations drop-down without having any translated content.

warioddly: I will add translations for the pages in the future.

PR #4146, opened 2024-04-15
chore: call `addError` by default on error in `Emitter`'s `forEach`

Gambley1:
## Status

READY

## Breaking Changes

NO

## Description

Add a default `addError` call to `BlocObserver` with error and stackTrace from `onError` function that comes from `Emitter`'s `forEach` function.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/emitter.dart
+++ b/packages/bloc/lib/src/emitter.dart
@@ -102,6 +102,7 @@ class _Emitter<State> implements Emitter<State> {
       onData: (data) => call(onData(data)),
       onError: onError != null
           ? (Object error, StackTrace stackTrace) {
+              addError(error, stackTrace);
               call(onError(error, stackTrace));
             }
           : null,


packages/bloc/lib/src/emitter.dart
```diff
@@ -102,6 +102,7 @@ class _Emitter<State> implements Emitter<State> {
       onData: (data) => call(onData(data)),
       onError: onError != null
           ? (Object error, StackTrace stackTrace) {
+              addError(error, stackTrace);
```
felangel, line 105: Can you provide some more context regarding why you want this change? Typically you use `addError` to notify an observer of an unhandled error. In This case, the error is handled because an `onError` callback is provided so I don't think it would be consistent to automatically call `addError` here.

felangel (changes requested):
Thanks for the PR. See my comment and lmk what you think. It would be great if you could file a detailed issue/feature request describing what you're trying to achieve so that we can discuss it before making code changes, thanks!

felangel:
Closing this PR for now. If you feel strongly about this change, please file an issue describing the problem/use-case you have so that we can align before opening a pull request, thanks!

PR #4157, opened 2024-04-27
refactor(flutter_todos): add `close` to `TodosApi`

### LINKED ISSUE #4154: question: [flutter_todos] why don't close StreamController?
**Description**

Hi, I'm learning about bloc-to-bloc communication through domain layer using streams. I was looking the example flutter_todos and I realised that local_storage_todos_api use an StreamController, but this is never closed.

This is not a bad practice? Is there some motive to not close this Stream Controller?

Thank you.

felangel:
## Status

**READY**

## Breaking Changes

NO

## Description

Even though the lifetime of the `TodosApi` instance is tied to the application in this case, it's still valuable to add a `close` API to the interface since the `TodosApi` should expose a way to close the instance and free any allocated resources. (closes #4154)

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/examples/flutter_todos/packages/local_storage_todos_api/lib/src/local_storage_todos_api.dart
+++ b/examples/flutter_todos/packages/local_storage_todos_api/lib/src/local_storage_todos_api.dart
@@ -19,7 +19,9 @@ class LocalStorageTodosApi extends TodosApi {
 
   final SharedPreferences _plugin;
 
-  final _todoStreamController = BehaviorSubject<List<Todo>>.seeded(const []);
+  late final _todoStreamController = BehaviorSubject<List<Todo>>.seeded(
+    const [],
+  );
 
   /// The key used for storing the todos locally.
   ///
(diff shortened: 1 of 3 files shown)


examples/flutter_todos/packages/local_storage_todos_api/lib/src/local_storage_todos_api.dart
```diff
@@ -19,7 +19,9 @@ class LocalStorageTodosApi extends TodosApi {
 
   final SharedPreferences _plugin;
 
-  final _todoStreamController = BehaviorSubject<List<Todo>>.seeded(const []);
+  late final _todoStreamController = BehaviorSubject<List<Todo>>.seeded(
```
felangel, line 22: Making this late so that the BehaviorSubject is instantiated lazily.

alfonsocejudo:
Hi there, quick question. When is the `local_storage_todos_api.dart` `Future<void> close()` method actually called? Is it automatically called by some process, thus it's enough to just implement and expose it? I wanted to play it safe so I figured I would add a `Future<void> close()` to my repository that will call the API instance's `close()`, and I then make this call to the repository's method from my Bloc's `@override Future<void> close()`, so when the BlocProvider needs cleaning up, the streams get cleaned up along with it. Thanks for your time, nice work.

Issue #4165, opened 2024-05-07
fix: emitting a state in a closed Cubit throws an error


ra48ad:
**Description**
Trying to emit a state when the cubit is closed (can happen for example in async functions) leads to an Error being thrown. This behaviour may be intentional, but it is - however - different from a Bloc's behaviour the emit is called.

**Steps To Reproduce**

1. Create a new Cubit
3. Add an async function that waits for 5 seconds, then emits a new state.
4. Create a new instance of this Cubit in your main method and call the function in 2) and then close the Cubit.
5. You should get an Error thrown that says "Cannot emit new states after calling close"

**Expected Behavior**
Calling emit on a closed Cubit ignores the new state and throws no error (Just like a Bloc would)

**Screenshots**

**Additional Context**

felangel:
Hi @ra48ad 👋 
Thanks for opening an issue!

Are you able to share a link to a minimal reproduction sample illustrating the problem/inconsistency you're seeing? It'd be much easier to help if you could share a repro, thanks!

ra48ad:
> Hi @ra48ad 👋 Thanks for opening an issue!
> 
> Are you able to share a link to a minimal reproduction sample illustrating the problem/inconsistency you're seeing? It'd be much easier to help if you could share a repro, thanks!

Thank you for your reply!

Sure, here is two comparable samples for a Cubit and a Bloc:

**Cubit**
```
class CounterCubit extends Cubit<int> {
  CounterCubit() : super(0);

  // delay emit to simulate async operation
  void increment() => Future.delayed(
        Duration(seconds: 3),
        () => emit(state + 1),
      );
}

Future<void> main(List<String> args) async {
  final cubit = CounterCubit();
  cubit.increment();
  await cubit.close();
  // prints 0, exception [StateError] thrown (bloc_base.dart line 97)
  print(cubit.state);
}
```

**Bloc**
```
enum CounterEvent { increment }

class CounterBloc extends Bloc<CounterEvent, int> {
  CounterBloc() : super(0) {
    on<CounterEvent>(_onCounterEvent);
  }

  Future<void> _onCounterEvent(CounterEvent event, Emitter<int> emit) async {
    if (event == CounterEvent.increment) {
      // delay emit to simulate async operation
      return Future.delayed(
        Duration(seconds: 3),
        () => emit(state + 1),
      );
    }
  }
}

void main(List<String> args) {
  final bloc = CounterBloc();
  bloc.add(CounterEvent.increment);
[...]

PR #4187, opened 2024-06-10
ignore emit in closed cubit

ra48ad:
## Status

**IN DEVELOPMENT**

## Breaking Changes

YES

## Description

This change will solve the inconsistency between Bloc and Cubit if emit is used after the Bloc/Cubit is closed. (Issue #4165)
The added emit override makes sure no error is thrown if emit is called and the cubit is already closed. Instead, the emit will be simply ignored just like a Bloc.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel (changes requested):
I don't think this is the right change. The bloc and cubit behavior is inconsistent because there is a slight delay when adding an event due to the Event stream/sink not because the `emit` behavior is different. If we are to adjust this it should likely be in `BlocBase`. I'm going to close this for now since I want to take some time to consider the implications of this change before proceeding.

PR #4193, opened 2024-06-18
make rethrow error optional

shivamkj:
## Status

**READY**

## Breaking Changes

NO

## Description

By default, bloc will rethrow any error caught during the event, which can't be configured. Now suppose, you have multiple events doing similar things (like making an API request), so you wrote a reusable error handler that can handle errors for all these events. But now you have to to wrap each of these event handlers in try catch and call the error handler in catch block of each event handler. If we make rethrow configurable, we can just override the  `onError` of the bloc and handle error from just single place. Note: You can still handle error in `onError` without adding try catch but Dart will complain about an uncaught error for every error throw, even though you have handled it in `onError`. This PR tries to make that rethrow optional with a condition that end developers can configure as per their needs. By default condition is true, so it will work as it works now.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -47,6 +47,9 @@ abstract class Bloc<Event, State> extends BlocBase<State>
   /// The current [BlocObserver] instance.
   static BlocObserver observer = const _DefaultBlocObserver();
 
+  /// weather to rethrow any error occured on event, defaults to true
+  static bool rethrowError = true;
+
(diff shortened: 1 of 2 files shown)


felangel:
Can you please file an issue with a minimal reproduction sample that illustrates the issue before opening a PR? That way I can better understand the issue and we can align on a solution before writing any code. Thanks!

PR #4207, opened 2024-07-20
A new widget block: BlocMuteSelector. Select a part of the state and create a mutable object.

MattiaPispisa:
A new widget block: BlocMuteSelector. Based on the state of a bloc, it allows you to select a part of the state and create a mutable object.

## Status

**READY**

## Breaking Changes

NO

## Description

A use case I encounter in some situations (especially when creating complex forms) is having an object that changes over time but should not reside in the bloc state. For example, I have TextFormFields that require using a TextEditingController because their content can change due to actions outside the field itself. Recreating the TextEditingController would necessitate repositioning the cursor correctly and dealing with other issues. By creating a mutable state, I can create and modify a TextEditingController. The `muteOrCopy` function takes the previous mutable state and the current bloc state as inputs, allowing me to apply controls based on the previous value. Since it is mutable, you cannot rely on the `muteOrCopy` function to execute the `builder` only when the state has actually changed, so `muteWhen` is used, which effectively works like BlocListener's listenWhen. The `muteOrCopy` function has a dual purpose; if it returns null, the state is expected to be mutable, but if it returns a new state, a comparison between the previous and new state is applied to update only if it has actually changed. The `create` function is executed on the first render, while the `muteOrCopy` function is applied for all subsequent renders.

Another use case where this can be helpful is when ValueNotifiers are required to satisfy third-party libraries, and I do not want to change the bloc state based on new third-party constraints.

## Type of Change

- [ X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🧹 Code refactor
- [ X] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/README.md
+++ b/packages/flutter_bloc/README.md
@@ -188,6 +188,26 @@ BlocSelector<BlocA, BlocAState, SelectedState>(
 )
 ```
 
+### BlocMuteSelector
+
+**BlocMuteSelector** is a Flutter widget which is analogous to `BlocBuilder` but allows developers to filter updates by selecting and mute a new value based on the current bloc state. Unnecessary builds are prevented using muteWhen. The selected value can be mutable and can mute using the previous value.
+
+If the `bloc` parameter is omitted, `BlocMuteSelector` will automatically perform a lookup using `BlocProvider` and the current `BuildContext`.
+
+```dart
+BlocMuteSelector<BlocA, BlocAState, MutableSelectedState>(
+   create: (state) {
+     // create the mutable state
+   }
+   muteOrCopy: (prev, state) {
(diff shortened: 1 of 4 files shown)


felangel:
@MattiaPispisa can you please file an issue clearly describing the problem you’re facing so we can discuss and align on a solution? Thanks!

MattiaPispisa:
Sure, I created the issue under "Feature Request" to discuss my problem, as it seemed like the most appropriate category among the available ones.

[Issue](https://github.com/felangel/bloc/issues/4209)

felangel:
Closing this for now while we discuss the solution in the issue. I'll follow up this weekend, thanks!

PR #4225, opened 2024-08-04
fix(flutter_todos): fix zone mismatch error

LukasMirbt:
## Status

**READY**

## Breaking Changes

NO

## Description

Fixes zone mismatch error by removing `runZonedGuarded` and handling errors through `PlatformDispatcher` instead.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/examples/flutter_todos/lib/bootstrap.dart
+++ b/examples/flutter_todos/lib/bootstrap.dart
@@ -1,7 +1,7 @@
-import 'dart:async';
 import 'dart:developer';
 
 import 'package:bloc/bloc.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_todos/app/app.dart';
 import 'package:flutter_todos/app/app_bloc_observer.dart';
@@ -13,12 +13,14 @@ void bootstrap({required TodosApi todosApi}) {
     log(details.exceptionAsString(), stackTrace: details.stack);
   };
 
+  PlatformDispatcher.instance.onError = (error, stack) {
+    log(error.toString(), stackTrace: stack);
+    return true;
+  };
+
   Bloc.observer = const AppBlocObserver();
 
   final todosRepository = TodosRepository(todosApi: todosApi);
 
-  runZonedGuarded(
-    () => runApp(App(todosRepository: todosRepository)),
-    (error, stackTrace) => log(error.toString(), stackTrace: stackTrace),
-  );
+  runApp(App(todosRepository: todosRepository));
 }


examples/flutter_todos/lib/bootstrap.dart
```diff
@@ -13,12 +13,14 @@ void bootstrap({required TodosApi todosApi}) {
     log(details.exceptionAsString(), stackTrace: details.stack);
   };
 
+  PlatformDispatcher.instance.onError = (error, stack) {
```
felangel, line 16: Can you elaborate a bit regarding the advantage of using PlatformDispatcher instead of wrapping `runApp` in a guarded zone? 

Also we should probably using the `WidgetsBinding` to access the platform dispatcher instance as per:
> Consider avoiding static references to this singleton through [PlatformDispatcher.instance](https://api.flutter.dev/flutter/dart-ui/PlatformDispatcher/instance.html) and instead prefer using a binding for dependency resolution such as WidgetsBinding.instance.platformDispatcher. See [PlatformDispatcher.instance](https://api.flutter.dev/flutter/dart-ui/PlatformDispatcher/instance.html) for more information about why this is preferred.

LukasMirbt: Hi!

Yea, sure!

The reason I opened this PR is because I got the following error in the console when running the `flutter_todo` example:

> Zone mismatch.
>       The Flutter bindings were initialized in a different zone than is now being used. This will likely cause confusion and bugs as any zone-specific configuration will inconsistently use the configuration of the original binding initialization zone or this zone based on hard-to-predict factors such as which zone was active when a particular callback was set.
>       It is important to use the same zone when calling `ensureInitialized` on the binding as when calling `runApp` later.
>       To make this warning fatal, set BindingBase.debugZoneErrorsAreFatal to true before the bindings are initialized (i.e. as the first statement in `void main() { }`).

The migration guide for this error in the [Flutter documentation
](https://docs.flutter.dev/release/breaking-changes/zone-errors#migration-guide) suggests that
> The best way to silence this message is to remove use of Zones from within the application.
 
(I also couldn't find any way to make the error go away and keep `runZonedGuarded`).

> Also we should probably using the WidgetsBinding to access the platform dispatcher instance as per:

I based the PR on the following snippet from the [Handling errors in Flutter](https://docs.flutter.dev/testing/errors#errors-not-caught-by-flutter) documentation:

```
import 'package:flutter/material.dart';
import 'dart:ui';

void main() {
  MyBackend myBackend = MyBackend();
  PlatformDispatcher.instance.onError = (error, stack) {
    myBackend.sendError(error, stack);
    return true;
  };
  runApp(const MyApp());
}
```

> Consider avoiding static references to this singleton through [PlatformDispatcher.instance](https://api.flutter.dev/flutter/dart-ui/PlatformDispatcher/instance.html) and instead prefer using a binding for dependency resolution such as WidgetsBinding.instance.platformDispatcher. See [PlatformDispatcher.instance](https://api.flutter.dev/flutter/dart-ui/PlatformDispatcher/instance.html) for more information about why this is preferred.

The same documentation mentions that
> The only place that WidgetsBinding.instance.platformDispatcher is inappropriate is if access to these APIs is required before the binding is initialized by invoking runApp() or WidgetsFlutterBinding.instance.ensureInitialized(). In that case, it is necessary (though unfortunate) to use the [PlatformDispatcher.instance](https://api.flutter.dev/flutter/dart-ui/PlatformDispatcher/instance.html) object statically.

so it might make sense to use PlatformDispatcher.instance directly if we don't also want to add `WidgetsFlutterBinding.instance.ensureInitialized()`.

felangel: > The reason I opened this PR is because I got the following error in the console when running the flutter_todos example

You just ran the example with no changes and saw the error? I'll try to see if I can reproduce, thanks for the context!

LukasMirbt: Can confirm that I'm still seeing this error on Flutter 3.24.1 on multiple platforms

felangel: Thanks I'll review this later today, apologies for the delay!

LukasMirbt: No rush!

Would be very interested to hear your thoughts on #4229 also whenever you have time 🚀

LukasMirbt: > I ran pub upgrade to resolve the win32 issue but don't see any Zone mismatch error 🤔

Strange 🤔

> It's fine either way I'll get this merged shortly, thanks!

Alright, no problem! 👍

felangel (approved):

PR #4268, opened 2024-10-26
feat: create debounce and debounce first transformers

mrgnhnt96:
## Status

**READY**

## Breaking Changes

NO

## Description

This PR introduces a new transformer:

### Debounce

This is the traditional debounce transformer. When any event is received, a "quiet period" will be created, based on the `duration` provided. When the window completes, the last event received will be processed.

```plaintext
# duration = 100ms
Event 1 -- 10ms -- Event 2 -- 10ms -- Event 3
```
- Event 1: debounce
- Event 2: debounce
- Event 3: process

#### Leading

When `leading = true`, then the first event will not be debounced, rather processed. The "quiet period" will still be created and the last received event will still be processed.

```plaintext
# duration = 100ms
Event 1 -- 10ms -- Event 2 -- 10ms -- Event 3
```

- Event 1: process
- Event 2: debounce
- Event 3: process

#### Transformer

After the "quiet period" is complete, it's possible that the previous event hasn't resolved yet, necessitating a second transformer to handle the events to process. The `concurrent` transformer is the default here.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/bloc_concurrency/lib/src/debounce.dart
```diff
@@ -0,0 +1,63 @@
+import 'dart:async';
+
+import 'package:bloc/bloc.dart';
+
+/// {@macro debounce}
+EventTransformer<E> debounce<E>([
+  Duration duration = const Duration(milliseconds: 300),
```
felangel, line 7: I think it'd be nice to have the duration be a required, named argument and to align with the API from package:stream_transform. In fact, we already depend on package:stream_transform so I think we should be able to dramatically simplify the implementation here.

I'm thinking the method signature could look something like:

```dart
EventTransformer<E> debounce<E>({
  required Duration duration,
  bool leading = false,
  bool trailing = true,
}) {...}
```

This would also allow us to remove `debounceFirst` altogether and there'd just be a single `debounce` transformer.

felangel: I'm also just remembering why I never included `debounce` to begin with -- because it's purely rate limiting the event stream but doesn't indicate how the emitted events are processed (e.g. are they still processed concurrently, sequentially, etc?).

felangel:
Left a comment based on my recollection and I think it'd be helpful to better understand the problem we're solving here. Can you maybe describe some real-world use-cases that would benefit from a `debounce` transformer where `droppable` and/or `restartable` won't help? Thanks!

mrgnhnt96:
Thank you for the feedback!

The main real-world use-case for a `debounce` transformer is to handle user interactions that trigger frequent events, such as search inputs or scrolling. These situations can quickly flood the event stream, and `debounce` helps limit the number of events processed. Specifically, `droppable` and `restartable` might not be suitable here because they either drop intermediate events or restart processing, while `debounce` simply defers processing until a certain “window period" is reached. `Debounce` would mean that only the final value after a burst of events is processed, without dropping or restarting events.

I see your point that `debounce` could already be implemented using the existing transformers along with `stream_transform`. The goal with this PR was to simplify the API and provide a ready-made, cohesive solution directly within `bloc_concurrency`, making it more accessible for developers who may want this common behavior without having to build their own transformer each time.

Thanks for suggesting the changes to the method signature, like making `duration` required and matching the `stream_transform` API. It makes sense to simplify things by using `stream_transform`, so I'll update it.

Regarding the concern about how the events should be processed (concurrently, sequentially, etc.), I think it might be best to start with a simple solution and expand on it in the future if the community's needs dictate.

mrgnhnt96:
Just pushed my changes, wanted to note on a few things:

1. I used `concurrent` as the default transformer to handle events to be processed within the `debounce` transformer. I felt this was fitting as the default transformer for bloc is `concurrent`

2. I didn't include the `trailing` parameter in the signature for `debounce`. This is because if `leading = true` and `trailing = false`, then essentially we've recreated the `droppable` transformer. LMK if you think that we should still include this and I'll revert [this commit](https://github.com/felangel/bloc/pull/4268/commits/e0fdf0a3527d482fcd4e79c166a2ac6189ed3f81)

felangel:
> Just pushed my changes, wanted to note on a few things:
> 
> 1. I used `concurrent` as the default transformer to handle events to be processed within the `debounce` transformer. I felt this was fitting as the default transformer for bloc is `concurrent`
> 2. I didn't include the `trailing` parameter in the signature for `debounce`. This is because if `leading = true` and `trailing = false`, then essentially we've recreated the `droppable` transformer. LMK if you think that we should still include this and I'll revert [this commit](https://github.com/felangel/bloc/pull/4268/commits/e0fdf0a3527d482fcd4e79c166a2ac6189ed3f81)

Thanks for all the hard work. The problem I'm having is, as I mentioned, `debounce` is not really an event transformer.

An event transformer = an optional event filter + a required event mapper.

Debounce is just an event filter (it doesn't imply anything about how the debounced events will be processed). In addition, it's trivial for folks to create their own custom debouncing transformer:

```dart
import 'package:bloc_concurrency/bloc_concurrency.dart';
import 'package:stream_transform/stream_transform.dart';

EventTransformer<E> concurrentDebounce<E>({
  required Duration duration,
  bool leading = false,
  bool trailing = true,
}) {
  return (events, mapper) {
    return concurrent<E>().call(
      events.debounce(duration, leading: leading, trailing: trailing),
      mapper,
    );
  };
}

EventTransformer<E> sequentialDebounce<E>({
  required Duration duration,
  bool leading = false,
  bool trailing = true,
}) {
  return (events, mapper) {
    return sequential<E>().call(
      events.debounce(duration, leading: leading, trailing: trailing),
      mapper,
    );
  };
}
```

It seems like the bigger issue here is the lack of advanced docs on writing event transformers which I am happy to work on.

For now though, I don't think it makes sense for a `debounce` event transformer to be included because by definition `debounce` is only part of an event transformer (without a mapper it's just a filter).

Hope that makes sense and sorry for leading you down this rabbit hole -- I totally forgot why I didn't do this in the first place!

Closing for now but happy to continue the discussion 👍

PR #4293, opened 2024-11-30
feat(bloc)!: add `emit.onEach` and `emit.forEach` to `Cubit`

### LINKED ISSUE #3555: feat: Add emitForEach / emitOnEach Methods (or something similar) to Cubit

felangel:
## Status

**IN DEVELOPMENT**

## Breaking Changes

YES

## Description

- feat(bloc)!: add `emit.onEach` and `emit.forEach` to `Cubit` (closes #3555)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
After thinking about this some more and chatting with more folks I don't feel this change provides a ton of value since the functionality in this implementation cancels any pending streams whenever `emit.forEach`/`emit.onEach` is used (regardless of the method it's used in). This differs from the bloc implementation since `emit.forEach`/`emit.onEach` is scoped to the event handler so `emit.forEach` in event handler A won't cancel the subscription for `emit.forEach` in event handler B. This isn't possible to due with Cubit (at least not with the current public API) and I don't think this change is worth it since I fear it won't solve real problems folks have and it'll lead to confusion since the API is named the same but functions differently across bloc and cubit.

PR #4332, opened 2025-01-13
Added to allow setting the validity period of hydrated data.

devstar1224:
## Status

**READY**

## Breaking Changes

NO

## Description

- Added to allow setting the validity period of hydrated data.

## Example
```dart
class CounterCubit extends HydratedCubit<int> {
  CounterCubit() : super(0, storage: EncryptedStorage());

  @override
  int fromJson(Map<String, dynamic> json) => json['value'] as int;

  @override
  Map<String, int> toJson(int state) => { 'value': state };

  @override
  Duration? get hydrationExpiresIn => const Duration(minutes: 1);
}
```

## Type of Change

- [ x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/hydrated_bloc/README.md
+++ b/packages/hydrated_bloc/README.md
@@ -145,6 +145,27 @@ class CounterCubit extends HydratedCubit<int> {
 }
 ```
 
+### Cubit storage set expireIn
+
+You can override the `hydrationExpiresIn` to customize the expiration period of a specific hydration using instances of `HydratedBloc` or `HydratedCubit`.
+
+```dart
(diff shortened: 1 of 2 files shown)


felangel:
Hi @devstar1224 👋 
Thanks for opening a PR!

I think it would be better if we could close this PR for now and move to an issue where we can discuss the problem being solved, align on a solution, and then open a pull request. Thanks so much! 🙏

PR #4348, opened 2025-01-30
feat(vscode): add useUnionTypes setting

zbarbuto:
## Status

**READY**

## Breaking Changes

NO

## Description

Adds a new `useUnionTypes` setting to both `newBlocTemplate` and `newCubitTemplate`. Defaults to `true` to backwards compatibility. When turned off, state classes will be generated as a single freezed class rather than using unions.

Also brings support for the existing `useSealedClasses` to freezed state generation. I'm happy to remove this from the change set if you'd prefer not changing two things in the same PR but it did seem harmless enough...

Example class generated with the setting as `true` (the default)

```
part of 'foo_cubit.dart';

@freezed
sealed class FooState with _$FooState {
  const factory FooState.initial() = _Initial;
}
```

Example class generated with the setting as `false` (the default)

```
part of 'foo_bloc.dart';

@freezed
sealed class FooState with _$FooState {
  const factory FooState() = _FooState;
}
```

This is useful for states that can rely solely on the `copyWith` generated from freezed to emit state changes. For example:

```
@freezed
sealed class CounterState with _$CounterState {
  const factory CounterState({ required int count }) = _CounterState;
}
```

No need for an initial state here and for changes we can just use `emit(state.copyWith( count: state.count + 1 ))`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -68,6 +68,11 @@
             "default": true,
             "description": "Whether to use sealed classes when creating a new bloc."
           },
+          "bloc.newBlocTemplate.useUnionTypes": {
+            "type": "boolean",
+            "default": true,
+            "description": "Whether to use union types for state or a single class relying on copyWith. Only works with \"freezed\"."
+          },
           "bloc.newCubitTemplate.type": {
             "type": "string",
             "default": "auto",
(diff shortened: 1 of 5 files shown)


zbarbuto:
Thanks for taking a look @felangel 

So basically all the templates like this:

```
@immutable
${classPrefix} class ${pascalCaseBlocName}State {}

${subclassPrefix}class ${pascalCaseBlocName}Initial extends ${pascalCaseBlocName}State {}
```

should omit the `Initial` version class if `useSealedClasses` is off?

tenhobi:
Not sure what is the default now, but I vote for making Dart sealed classes as default from now on.

Issue #4136, opened 2024-04-02
feat: Add `force` to Emitter


mcmah309:
**Description**

Most of the time, the current bloc implementation does exactly what you would want, only emit changes if the state has actually changed. Which is accomplished through the `==` operator. This seems to work fine until you get to large complex states, then this becomes a performance/usability bottleneck. 

e.g. I currently have ephemeral virtual file system structures inside my bloc's state. Each time the file system changes, it would be ridiculous to copy the entire virtual file system structure.

Currently when you want to signal that state has changed, you would do
```dart
emit(newState);
```
To make this work with a virtual file system, as mentioned, I would have to copy the entire file system and possibly write traversing logic in the `==` operator determine if my state as changed. To resolve these scenarios, a function like
```dart
emit.force(newState);
```
or
```dart
emit(newState, force: true)
```
should be added.

**Desired Solution**

The `Emitter` class should have a `force(...)` method or `call` have an optional `force` parameter to signal the state has updated, without needing to do equality comparison.

**Alternatives Considered**

One could override the `==` operator on the state object to always return `false`, but this is not a desired solution because this would be abuse of `==`, since `==` would return `false`, indicating that the objects are different, but it does not actually mean they are different. This would be leaking emitter logic into the state and making use of `==` anywhere else in different contexts (like a hashmap of file system nodes) yield wrong results.

tenhobi:
I just feel like you can just use Equatable for all objects of your state representation and then you can actually just emit the "same state" (mutated), because now it will be recognized as new state?

mcmah309:
- It would be a very inefficient solution to traverse an entire file system to determine equality.
- For an object, you may not want to track every field in `==` or `hashcode`. e.g. An id field of an object is used for fast retrieval/updating cache of correct object if fields change.
- `==` or `hashcode` may not represent pure equality of all the fields. e.g. data structures that allow hashing different types in the same hash table or using keys of one type to retrieve another type.
- Objects in a state may be outside your library to do not properly override `==`.
- The state object may contain objects significantly large. e.g. an entire document, copying would be inefficient.

There are a lot of reasons for allowing a `force` param. The framework shouldn't pigeon hole developers into a you must copy and compare, if you don't agree you have to leak emitter logic into the `==` operator of the state or switch to a different framework for this one scenario. Let the developer decide for the scenario. I am always in favor of giving developers more control. Especially when a solution like adding `force` to emit has zero overhead for the framework or effect on anyone who chooses not to use it.

[...]

PR #4372, opened 2025-03-11
feat(bloc_selector): BlocSelector only rebuilds when selected collection changes

nfdz:
## Status

**READY**

## Breaking Changes

NO

## Description

Dart collections, for example List, are only equal to themselves by default. [Dart API](https://api.flutter.dev/flutter/dart-core/List/operator_equals.html).

When using **BlocSelector** and the selected type is a **collection**, then we cannot guarantee the correct functioning of BlocSelector, and **the builder will be invoked every time the state of the Bloc changes**.

This scenario is very common. I have often seen in various projects a List being selected inside a state that contains many other variables.

This solution tries to be as simple as possible, reusing the official solution provided by Dart, for example [listEquals](https://api.flutter.dev/flutter/foundation/listEquals.html).

The proposed solution:
```dart
BlocSelector<BlocA, BlocAState, List<EntryState>>(
   selector: (state) { ... },
   builder: (context, state) { ... },
   selectorEquality: listEquals,
)
```

## Type of Change

- [ x ] ✨ New feature (non-breaking change which adds functionality)
- [ x ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_selector.dart
+++ b/packages/flutter_bloc/lib/src/bloc_selector.dart
@@ -32,6 +32,7 @@ class BlocSelector<B extends StateStreamable<S>, S, T> extends StatefulWidget {
     required this.builder,
     Key? key,
     this.bloc,
+    this.selectorEquality,
   }) : super(key: key);
 
   /// The [bloc] that the [BlocSelector] will interact with.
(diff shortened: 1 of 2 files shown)


felangel:
Thanks for opening a PR!

I'd really prefer not to introduce something like this because of the reasons described in https://github.com/felangel/bloc/issues/4136#issuecomment-2081169797.

If you still feel strongly please file an issue clearly describing your use case and I'm more than happy to discuss further 👍

nfdz:
No problem, I understand your point of view.

I see that this is very easy to fix from a Bloc dev-user's perspective. But even so, it's something I've seen too many times, and I wanted to share my concern. 
I hope future major versions of Bloc fix this issue with an small breaking change and make it possible to use BlocSelector with collections. I prefer that to the workaround of this PR because I don't see any use-case for using BlocSelector with collections like it is today.

Thank you for your amazing work!

Issue #4383, opened 2025-03-26
fix: MultiProvider error when attempting to build


kupresk14:
**It is the Provider package, override to 6.1.2 for now to fix temporarily like @esteban-farias-ueno mentions in the comments. They updated about an hour ago from this edit** 

**Description**
When using flutter_bloc 8.1.6, and even after attempting an update to 9.0.0 we run into an error:

`Superclass has no constructor named 'MultiProvider'`

We are on Flutter 3.24.3, so a bit behind

**Steps To Reproduce**

1. Attempt to build an Android APK
2. See that a MultiProvider error pops up
3. Get confused and investigate

**Expected Behavior**
We should be able to build without the error showing up.

**Screenshots**
Your own static analysis section is failing:

<img width="779" alt="Image" src="https://github.com/user-attachments/assets/5c5410fa-84c1-4921-be1f-c4d784217c10" />

Within our Actions:
![Image](https://github.com/user-attachments/assets/abf91307-6489-44ad-aa8b-5c5b5a7ed238)

**Additional Context**
It seems this only started happening in the last few hours, we had no issues before that.

esteban-farias-ueno:
for now just make an override of this dedepndency: provider: 6.1.2

kupresk14:
> for now just make an override of this dedepndency: provider: 6.1.2

Thank you!

m-1226:
can you help me the same issue ?

esteban-farias-ueno:
> can you help me the same issue ?

in your pusbpec.yaml  add this:
dependency_override:
  provider: 6.1.2

then flutter pub upgrade.
Your test should work now.

lampian:
Please help, started picking up this issue a couple of hours ago. 

Running on 3.19.2
deps:
flutter_bloc on ^9.1.0. (also picking it up on ^8.1.6
provider on ^6.1.2

Did the rep override. 

..need to provide working build in 10hours  - assistance will be greatly appreciated!

esteban-farias-ueno:
> Please help, started picking up this issue a couple of hours ago.
> 
> Running on 3.19.2 deps: flutter_bloc on ^9.1.0. (also picking it up on ^8.1.6 provider on ^6.1.2
> 
> Did the rep override.
> 
> ..need to provide working build in 10hours - assistance will be greatly appreciated!

exactly like this ?
dependency_override:
 provider: 6.1.2

flutter pub upgrade?? try with flutter clean and then flutter pub get

[...]

PR #4384, opened 2025-03-26
fix(flutter_bloc): support breaking change in `provider` v6.1.3

felangel:
## Status

**READY**

## Breaking Changes

YES

## Description

- refactor(flutter_bloc): support breaking change in `provider` v6.1.3
  - this is technically a breaking change because `MultiBlocProvider`, `MultiBlocListener`, and `MultiRepositoryProvider` are no longer subclasses of `MultiProvider` but we're forced to do so due to the breaking changes introduced in `pkg:provider` v6.1.3 (closes https://github.com/felangel/bloc/issues/4383 and related to https://github.com/rrousselGit/provider/issues/907).
  - I would have much preferred for @rrousselgit to have reverted the change and published 6.1.4 but it's been 5 hours since the breaking change was published and I don't want to keep waiting for an upstream fix since there hasn't been any update or indication that the change will be reverted and this is very disruptive change for all folks using `pkg:flutter_bloc`.
  

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/lib/src/multi_bloc_listener.dart
+++ b/packages/flutter_bloc/lib/src/multi_bloc_listener.dart
@@ -51,22 +51,21 @@ import 'package:provider/single_child_widget.dart';
 class MultiBlocListener extends StatelessWidget {
   /// {@macro multi_bloc_listener}
   const MultiBlocListener({
-    required this.listeners,
-    required this.child,
+    required List<SingleChildWidget> listeners,
+    required Widget child,
     Key? key,
-  }) : super(key: key);
+  })  : _listeners = listeners,
+        _child = child,
+        super(key: key);
 
-  /// The list of [BlocListener] instances.
-  final List<SingleChildWidget> listeners;
-
-  /// The [Widget] that will be rendered as a child.
-  final Widget child;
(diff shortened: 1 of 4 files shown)


felangel:
This was fixed in provider v6.1.4 so this change was reverted.

rrousselGit:
> * I would have much preferred for @rrousselGit to have reverted the change and published 6.1.4 but it's been 5 hours since the breaking change was published and I don't want to keep waiting for an upstream fix since there hasn't been any update or indication that the change will be reverted and this is very disruptive change for all folks using `pkg:flutter_bloc`.

Gotta sleep you know :)

It's late night here. Publishing the package was my last task of the day.

felangel:
> > * I would have much preferred for @rrousselGit to have reverted the change and published 6.1.4 but it's been 5 hours since the breaking change was published and I don't want to keep waiting for an upstream fix since there hasn't been any update or indication that the change will be reverted and this is very disruptive change for all folks using `pkg:flutter_bloc`.
> 
> Gotta sleep you know :)
> 
> It's late night here. Publishing the package was my last task of the day.

It was perfect timing 😂

Issue #3069, opened 2021-12-16
feat(cubit): cancelable async operations


felangel:
**Description**

As a developer, I want to be able to await asynchronous operations within a cubit which are automatically canceled if a cubit is closed while the async operation is pending.

An example use-case is when using a Cubit to fetch some data asynchronously from a screen in which a user can go back.

```dart
static Route<MyScreen> route() {
  return MaterialPageRoute(
    builder: (context) => BlocProvider(
      create: (context) => MyCubit()..load(),
      child: const MyScreen(),
    ),
  );
}

enum MyState { idle, loading }

class MyCubit extends Cubit<MyState> {
  MyCubit() : super(MyState.idle);

  Future<void> load() async {
    emit(MyState.loading);
    await Future<void>.delayed(const Duration(seconds: 3));
    emit(MyState.idle);
  }
}
```

In this scenario, as soon as `MyScreen` is pushed onto the navigation stack, the `load()` method is called on a newly created instance of `MyCubit`. It's possible that the user might get tired of waiting and press the back button before `load()` has completed. In this case, the Future will still complete after the `MyCubit` has been closed and the subsequent `emit(MyState.idle)` will be evaluated which will result in a `StateError`:

`Unhandled Exception: Bad state: Cannot emit new states after calling close`

**Desired Solution**

It would be nice if cubit had a `cancelable` (open to naming suggestions) API which allowed developers to await asynchronous operations which would automatically be canceled if the cubit was closed.

```dart
enum MyState { idle, loading }

class MyCubit extends Cubit<MyState> {
  MyCubit() : super(MyState.idle);

  Future<void> load() async {
    emit(MyState.loading);
    await cancelable(() => Future<void>.delayed(const Duration(seconds: 3)));
    emit(MyState.idle);
  }
}
```

**Alternatives Considered**

- Developers could check if the Cubit instance has been closed after any asynchronous operations and abort:

```dart
enum MyState { idle, loading }

class MyCubit extends Cubit<MyState> {
  MyCubit() : super(MyState.idle);

  Future<void> load() async {
    emit(MyState.loading);
    await Future<void>.delayed(const Duration(seconds: 3));
    if (isClosed) return;
    emit(MyState.idle);
  }
}
```

- Developers could also use `CancelableOperation` and `CancelableCompleter` from `package:async` to maintain a list of cancelable operations internally which are manually canceled when the cubit is closed.

- emit could automatically ignore states after the instance has been closed (previous behavior)

**Additional Context**

[...]

PR #4386, opened 2025-03-28
feat(bloc): Introduce maybeEmit to safely emit states

rumax:
Introduced method to safely emit state. The method allows to always trigger "maybeEmit" without a need to check if bloc is closed or not. In case if bloc is closed, the event will be ignored;

Resolves: #3069

## Status

**READY**

## Breaking Changes

YES

## Type of Change

- [ ] ~~🛠️ Bug fix (non-breaking change which fixes an issue)~~
- [ ] ~~❌ Breaking change (fix or feature that would cause existing functionality to change)~~
- [ ] ~~🧹 Code refactor~~
- [ ] ~~✅ Build configuration change~~
- [ ] ~~📝 Documentation~~
- [ ] ~~🗑️ Chore~~

### DIFF
--- a/packages/bloc/lib/src/bloc_base.dart
+++ b/packages/bloc/lib/src/bloc_base.dart
@@ -33,6 +33,9 @@ abstract class Closable {
 abstract class Emittable<State extends Object?> {
   /// Emits a new [state].
   void emit(State state);
+
+  /// Safely emits a new [state] if the bloc is not closed.
+  void maybeEmit(State state);
 }
 
 /// A generic destination for errors.
@@ -109,6 +112,21 @@ abstract class BlocBase<State>
       rethrow;
     }
   }
+
+  /// Safely updates the [state] to the provided [state] if the block is not
+  /// closed.
+  /// [emit] does nothing if the [state] being emitted
+  /// is equal to the current [state].
+  ///
+  /// To allow for the possibility of notifying listeners of the initial state,
+  /// emitting a state which is equal to the initial state is allowed as long
+  /// as it is the first thing emitted by the instance.
+  ///
+  /// * Throws a [StateError] if the bloc is closed.
+  @protected
+  @visibleForTesting
+  @override
+  void maybeEmit(State state) => isClosed ? null : emit(state);
 
   /// Called whenever a [change] occurs with the given [change].
   /// A [change] occurs when a new `state` is emitted.
--- a/packages/bloc/test/bloc_test.dart
+++ b/packages/bloc/test/bloc_test.dart
@@ -1540,6 +1540,16 @@ void main() {
         await counterBloc.close();
       });
 
+      test('safely updates the state with maybeEmit', () async {
+        final counterBloc = CounterBloc();
+        unawaited(
+          expectLater(counterBloc.stream, emitsInOrder(const <int>[42])),
+        );
+        counterBloc.maybeEmit(42);
+        expect(counterBloc.state, 42);
+        await counterBloc.close();
+      });
+
       test(
           'throws StateError and triggers onError '
           'when bloc is closed', () async {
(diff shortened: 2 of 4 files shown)


felangel:
@rumax thanks for the PR but since it's trivial to add this API yourself as an extension without any changes to the library, I'd prefer to hold off on making this change. I'm still considering alternatives but in the meantime if you want this API just define the following extension in your own project:

```dart
extension CubitMaybeEmit<S> on Cubit<S> {
  @protected
  void maybeEmit(S state) {
    if (isClosed) {
      return;
    }

    // ignore: invalid_use_of_protected_member, invalid_use_of_visible_for_testing_member
    emit(state);
  }
}
```

I also want to add that in general I don't recommend ignoring when the instance is closed because you can potentially end up executing really expensive operations for no reason.

For example,

```dart
enum MyState { idle, loading, loadedA, loadedB }

class MyCubit extends Cubit<MyState> {
  MyCubit() : super(MyState.idle);

  Future<void> load() async {
    emit(MyState.loading);
    final resultA = await _expensiveOperationA();
    if (isClosed) return;
    emit(MyState.loadedA);
    final resultB = await _expensiveOperationB();
    if (isClosed) return;
    emit(MyState.loadedB);
  }
}
```

If you rewrite this using `maybeEmit`:

```dart
enum MyState { idle, loading, loadedA, loadedB }

class MyCubit extends Cubit<MyState> {
  MyCubit() : super(MyState.idle);

  Future<void> load() async {
    emit(MyState.loading);
    final resultA = await _expensiveOperationA();
    maybeEmit(MyState.loadedA);
    final resultB = await _expensiveOperationB();
    maybeEmit(MyState.loadedB);
  }
}
```

Then you will always perform expensive operation B (even in cases where the result doesn't matter).

I hope this helps illustrate my point, closing for now but happy to continue the discussion in #3069.

rumax:
@felangel Your reply looks exactly like [my comment](https://github.com/felangel/bloc/issues/3069#issuecomment-2330811666), and it was a workaround, so thanks, but I'm aware of this solution. I kindly disagree with it. Let me explain why:

1) My PR doesn't break or violate the current implementation; the host app can decide which case to use. The change only provides new functionality that will simplify a lot of code. Consider these cases:

```
Future<void> load() async {
    emit(MyState.loading);
    
    final resultA = await _expensiveOperationA();

    if (isClosed) return;
    
    emit(MyState.loadedB);
  }
  
  Future<void> save() async {
    emit(MyState.loading);
    
    final resultA = await _expensiveOperationA_2();

    if (isClosed) return;
    
    emit(MyState.savedB);
  }
  
  void getCachedData () { // This function is even without Future.     
    final resultA = nonAsyncOp();
    
    emit(resultA);
  }
  ```
  
  As you can see, you always need to write `if (isClosed) return;`. Isn't it [Duplicate Code](https://code-smells.com/dispensables/duplicate-code)? Compare it to:
  
  ```
Future<void> load() async {
    emit(MyState.loading);
    
    final resultA = await _expensiveOperationA();
    
    maybeEmit(MyState.loadedB);
  }
  
  Future<void> save() async {
    emit(MyState.loading);
    
    final resultA = await _expensiveOperationA_2();
    
    maybeEmit(MyState.savedB);
  }
  
  void getCachedData => maybeEmit(nonAsyncOp());
  
  ```
  
  2) You make this example
  
```
Future<void> load() async {
    emit(MyState.loading);
    final resultA = await _expensiveOperationA();
    maybeEmit(MyState.loadedA);
    final resultB = await _expensiveOperationB();
    maybeEmit(MyState.loadedB);
  }
```

But who restricts the user from adding `if (isClosed) return;` before _expensiveOperationB? It's super easy to fix by only adding `if (isClosed) return;` before the second await. You might think that the host app will forget about this, and you "force" developers to add this check; otherwise, emit throws an exception. This can result in code where everything is surrounded with try-catch blocks because emit throws, and any of _expensiveOperationA or _expensiveOperationB can throw. The "safe" code you provided will result in:

```
Future<void> load() async {
    emit(MyState.loading);
    try {
      final resultA = await _expensiveOperationA();
      if (isClosed) return;
      emit(MyState.loadedA);
      final resultB = await _expensiveOperationB();
      if (isClosed) return;
      emit(MyState.loadedB);
    } catch () {
      emit(MyState.emptyOrErrorState);
    }
  }
  ```
  
3) Your suggestion of using `extension CubitMaybeEmit<S> on Cubit<S>` is only a workaround and it is a bad practice to put it to the production code, because it violates 2 rules

> invalid_use_of_protected_member, invalid_use_of_visible_for_testing_member

And as a result another code smell?

> Accessing Protected Members: The emit method is protected, meaning it should only be accessed within the class or its subclasses. Accessing it directly in an extension can break encapsulation principles.

My PR could easily fix it. Otherwise, once developer introduces one ignore, the project will be filled with other ignores too.

PR #4390, opened 2025-04-02
feat(hydrated_bloc): add `onError` and  `HydrationErrorBehavior` to `hydrate`

### LINKED ISSUE #4382: fix: Hydrated bloc looses stored state when error on `fromJson`
**Description**
When an exception happens on the `fromJson` of a hydrated cubit/bloc, the stored data is lost and reset to the initial data.

**Steps To Reproduce**

Here is a simple reproducible code:

```dart
import 'package:hydrated_bloc/hydrated_bloc.dart';

class CountCubit extends HydratedCubit<int> {

  CountCubit() : super(0);

  void decrement() => emit(state - 1);

  @override
  int? fromJson(Map<String, dynamic> json) {
    // throw Exception('error');
    return json['data'];
  }

  @override
  Map<String, dynamic>? toJson(int state) {
    return {'data': state};
  }
}

void main() async {
  WidgetsFlutterBinding.ensureInitialized();
  HydratedBloc.storage = await HydratedStorage.build(
    storageDirectory:
        kIsWeb
            ? HydratedStorageDirectory.web
            : HydratedStorageDirectory((await getTemporaryDirectory()).path),
  );
  runApp(
    MaterialApp(
      home: Scaffold(
        body: BlocProvider(
          create: (_) => CountCubit(),

          child: BlocBuilder<CountCubit, int>(
            builder:
                (context, state) => Center(
                  child: Row(
                    children: [
                      ElevatedButton(
                        child: Text('Increment'),
                        onPressed: () => context.read<CountCubit>().increment(),
                      ),
                      Text('$state'),
                      ElevatedButton(
                        child: Text('Decrement'),
                        onPressed: () => context.read<CountCubit>().decrement(),
                      ),
                    ],
                  ),
                ),
          ),
        ),
      ),
    ),
  );
}
```
1- Run the app
2- Interact with it by incrementing and decrementing
3- Hot restart and the data should still be there
4- Uncomment the throw exception in the `fromJson` method and hot restart
5- No exceptions where throw and the state is back at the initial one
6- Comment/Remove the throw, and hot restart
7- See how the state were lost.

**Expected Behavior**
I am not sure if this is a bug or not, probably just a specific use case?

Either way, it would be nice to have a way to check if the stored state is valid or not, so we could show the user an error or something, so we don't lose the stored state.

In my project I did this work around:

Created this `LazyHydratedCubit` which don't instantly call the hydrate method on the constructor: 

```dart
abstract class LazyHydratedCubit<State> extends Cubit<State>
    with HydratedMixin<State> {
  LazyHydratedCubit(super.state);

  Map<String, dynamic> get storagedValue {
    final storage = HydratedBloc.storage;
    final json = storage.read(storageToken) as Map<dynamic, dynamic>?;
    final dynamic traversedJson = _traverseRead(json);
    return _cast<Map<String, dynamic>>(traversedJson) ?? <String, dynamic>{};
  }

  dynamic _traverseRead(dynamic value) {
    if (value is Map) {
      return value.map<String, dynamic>((dynamic key, dynamic value) {
        return MapEntry<String, dynamic>(
          _cast<String>(key) ?? '',
          _traverseRead(value),
        );
      });
    }
    if (value is List) {
      for (var i = 0; i < value.length; i++) {
        value[i] = _traverseRead(value[i]);
      }
    }
    return value;
  }

  T? _cast<T>(dynamic x) => x is T ? x : null;
}```

Then on the page I did something like this:

```dart
class _GamePageState extends State<GamePage> {
  late final gameCubit = GameCubit(
    shipyardRepository: context.read(),
  );

  bool hasStateError = false;

  @override
  void initState() {
    super.initState();

    final storagedValue = gameCubit.storagedValue;

    try {
      GameState.fromJson((storagedValue as Map).cast<String, dynamic>());
      gameCubit.hydrate();
    } catch (e, s) {
      if (kDebugMode) {
        // TODO(erickzanardo): Make a better way to save logs
        print('Error while loading game state\n$e\n$s');
      }
      hasStateError = true;
    }
  }

  @override
    // omitted
   }
}
```
Maybe having a way for the bloc to not hydrated automatically would be enough? for this?

**Additional Context**

Happy to work on a PR if you think this use case would bring value

erickzanardo:
## Status

**IN DEVELOPMENT**

## Breaking Changes

NO

## Description

Fixes #4382

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
Planning to wrap this up later today. I want to adjust the tests a bit and write docs but otherwise should be good to land 🎉

packages/hydrated_bloc/README.md
```diff
@@ -145,7 +165,7 @@ Any `storageDirectory` can be used when creating an instance of `HydratedStorage
 final storage = await HydratedStorage.build(
   storageDirectory: await getApplicationDocumentsDirectory(),
 );
-```
+````
```
felangel, line 168:
```suggestion
```
```

felangel:
@erickzanardo I think this is finally ready for review. Let me know what you think of the changes I made when you have a few mins, thanks! 🙏

erickzanardo:
@felangel it looks great! Thanks for taking the time to make the adjustments!

PR #4405, opened 2025-04-11
feat(flutter_bloc): support listener current state

### LINKED ISSUE #4347: feat: Optionally support current state in `BlocListener`
**Description**

Given `BlocListener` is intended to react to _changes_ in state, the current default makes sense: the listener is not called immediately with the current state ([as documented](https://bloclibrary.dev/flutter-bloc-concepts/#bloclistener)).

However, I've come across a few instances where it would be useful to have the **option** for a listener run once with the current state in addition to future transitions. Generally, instances where you want some feedback in the UI based on state which isn't applicable in a `builder` or using `context.select` - like snackbars or route changes.

**Alternatives Considered**

The current workaround for this is to make sure the widget is stateful and include an `initState()`. For example, let's say we have a page that is only visible to logged-in users. We want it to re-route users that are not currently logged in in addition to users who log out in future.

```dart
class AuthenticatedPage extends StatefulWidget {
  const AuthenticatedPage({super.key});

  @override
  State<AuthenticatedPage> createState() => _AuthenticatedPageState();
}

class _AuthenticatedPageState extends State<AuthenticatedPage> {
  @override
  void initState() {
    super.initState();
    // Handle the current bloc state
    if (!context.read<AuthBloc>().state.isLoggedIn) {
      _signout();
    }
  }

  @override
    return BlocListener<AuthBloc, AuthState>(
      // Handle changes to bloc state
      listenWhen: (previous, next) => previous.isLoggedIn && !next.isLoggedIn,
      listener: (context, state) {
        _signout();
      },
      child: MyAuthenticatedPageContent(),
    );
  }

  _signout() {
    context.router.replace(const UnauthenticatedRoute());
  }
}
```

Alternatively, we use something like `StreamBuilder` with a parameter like

```
stream: context.read<MyCubit>().stream.startWith(context.read<MyCubit>().state);
```

But this feels way hackier.

**Desired Solution**

It would clean up a few things to have some sort of optional flag (like `startWithCurrentState` but probably a better name) that will immediately call the listener once with the current state. With this, we could simplify our widget down to be stateless:

```dart
class AuthenticatedPage extends StatelessWidget {
  @override
    return BlocListener<AuthBloc, AuthState>(
      // Ensure the listener is checked immediately on creation with current state in addition to future transitions
      startWithCurrentState: true,
      listenWhen: (previous, next) => previous.isLoggedIn && !next.isLoggedIn,
      listener: (context, state) {
        context.router.replace(const UnauthenticatedRoute());
      },
      child: MyAuthenticatedPageContent(),
    );
  }
}
```

If it makes more sense this could be its own widget rather than a flag on `BlocListener`.

zbarbuto:
## Status

**READY**

## Breaking Changes

NO

## Description

Allows setting a `startWithCurrentState` flag on `BlocListener` which changes the behaviour to immediately call `listener` with the current `state` of the bloc before calling again with subsequent emissions.

This allows the view layer to react to whatever the current state is rather than only when it changes.

For example:

```dart
return BlocListener<AuthBloc, AuthState>(
  // Ensure the listener is checked immediately on creation with
  // current state in addition to future transitions.
  // If the user is somehow routed to this page when not logged in
  // they should be immediately kicked out, rather than only when
  // they transition from logged in to logged out.
  startWithCurrentState: true,
  listener: (context, state) {
    if(!state.isLoggedIn) {
      context.router.replace(const UnauthenticatedRoute());
    }
  },
  child: MyAuthenticatedPageContent(),
);
```

Closes #4347

I don't love the name of the flag itself but I couldn't come up with anything better (naming things, amirite?).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/flutter_bloc/lib/src/bloc_listener.dart
+++ b/packages/flutter_bloc/lib/src/bloc_listener.dart
@@ -74,6 +74,31 @@ typedef BlocListenerCondition<S> = bool Function(S previous, S current);
 /// )
 /// ```
 /// {@endtemplate}
+///
+/// {@template bloc_listener_start_with_current_state}
+/// By default, the [listener] will only be called on new state emissions from
+/// the bloc. If you want to call the listener with the current state
+/// immediately, set [startWithCurrentState] to true.
+/// Note that [listenWhen] is not called for the current state (as there is no
+/// previous state to compare to).
+///
+/// ```dart
+/// // Default behavior
+/// BlocListener<BlocA, BlocAState>(
+///   listener: (context, state) {
+///     // this is only called when BlocA emits new states
+///   },
+/// )
+///
+/// BlocListener<BlocA, BlocAState>(
+///   startWithCurrentState: true,
+///   listener: (context, state) {
+///     // this is now called immediately with BlocA's current state
+///     // in addition to new emissions
+///   },
+/// )
+/// ```
+/// {@endtemplate}
 class BlocListener<B extends StateStreamable<S>, S>
     extends BlocListenerBase<B, S> {
   /// {@macro bloc_listener}
@@ -83,6 +108,7 @@ class BlocListener<B extends StateStreamable<S>, S>
     Key? key,
     B? bloc,
     BlocListenerCondition<S>? listenWhen,
+    bool startWithCurrentState = false,
     Widget? child,
   }) : super(
           key: key,
@@ -90,6 +116,7 @@ class BlocListener<B extends StateStreamable<S>, S>
           listener: listener,
           bloc: bloc,
           listenWhen: listenWhen,
+          startWithCurrentState: startWithCurrentState,
         );
 }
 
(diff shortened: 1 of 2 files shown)


nvshah:
@zbarbuto 

With this, I guess it will help us avoid writing boiler-plate ie checks in initialState (just to check on last emitted state !!)
Am I interpreting this change rightly ?

Then this indeed would be helpful +1 !!

zbarbuto:
@nvshah Yep that's pretty much the exact goal.

@felangel really sorry to ping you but do you have any feedback on this?

felangel:
> @nvshah Yep that's pretty much the exact goal.
> 
> @felangel really sorry to ping you but do you have any feedback on this?

Sorry for the delay! Will give this some more thought this weekend.

tenhobi:
I like this idea! In few "rare" cases I also needed or had issues with this, since UI was constructed in some cases later than bloc emited the state and there wasnt better way to handle this to make it stable. So I think either this flag or alternatively having `onAttach` callback described here https://github.com/felangel/bloc/issues/4495 would be handy.

I think I like this flag a bit more.

packages/flutter_bloc/lib/src/bloc_listener.dart
ryanaltair: if _bloc was replaced but with same state, 
this may trigger listener function.
Should we add config to control this behaviour?

PR #4415, opened 2025-04-18
fix(flutter_bloc): rebuild `BlocSelector` when `selector` changes

### LINKED ISSUE #4414: BlocSelector fails to rebuild on selector change, causing outdated state
### Description

When using BlocSelector, updating the selector logic dynamically does not trigger a rebuild or reflect the new selection criteria immediately. It appears that BlocSelector continues using the old selector function until a new event is emitted in the associated bloc.

This leads to stale state being shown, even though the selector has changed — as long as the bloc instance remains the same and no new event is dispatched.

---

### 🔁 Steps to Reproduce

1. Create a widget using `BlocSelector` with a dynamic `selector` function. (ie where selector function logic may vary)
2. Inside a `StatefulWidget`, toggle the selector logic using `setState`.
3. Keep the bloc instance the same.
4. Notice that `BlocSelector` continues using the **previous selector logic** until a new event is emitted to the bloc.

#### Minimal Example

```dart
import 'package:flutter/material.dart';
import 'package:flutter_bloc/flutter_bloc.dart';

class CounterCubit extends Cubit<int> {
  CounterCubit() : super(0);
}

class SelectorDemoPage extends StatelessWidget {
  const SelectorDemoPage({super.key});

  @override
    return BlocProvider(
      create: (_) => CounterCubit(),
      child: Scaffold(
        body: Center(
          child: Container(height: 300, width: 500, child: SelectorDemo()),
        ),
      ),
    );
  }
}

class SelectorDemo extends StatefulWidget {
  @override
  _SelectorDemoState createState() => _SelectorDemoState();
}

class _SelectorDemoState extends State<SelectorDemo> {
  bool useEvenSelector = true;

  int evenSelector(int state) => state.isEven ? state : -1;
  int oddSelector(int state) => state.isOdd ? state : -1;

  @override
    return Column(
      mainAxisAlignment: MainAxisAlignment.spaceBetween,
      crossAxisAlignment: CrossAxisAlignment.center,
      children: [
        ElevatedButton(
          onPressed: () {
            setState(() {
              useEvenSelector = !useEvenSelector;
            });
          },
          child: Text("Toggle Selector"),
        ),
        Text('Current Selector: ${useEvenSelector ? 'Even' : 'Odd'}'),
        ElevatedButton(
          onPressed: () {
            context.read<CounterCubit>().increment();
          },
          child: Text("Increment"),
        ),
        BlocSelector<CounterCubit, int, int>(
          selector:
              (state) =>
                  useEvenSelector ? evenSelector(state) : oddSelector(state),
          builder: (context, selected) {
            return Text(
              "Selected: $selected || (Actual Value: ${context.read<CounterCubit>().state}) > ",
            );
          },
        ),
      ],
    );
  }
}
```
---

#### 🧠 Code Explanation

- A `CounterCubit` emits incremented integers.
- Inside a `StatefulWidget`, we toggle between two selector functions:
  - `evenSelector` → returns the value only if it's even, else `-1`
  - `oddSelector` → returns the value only if it's odd, else `-1`
- `BlocSelector` uses `useEvenSelector` to decide which logic to apply, and that flag is toggled using `setState`.

---

#### ❗️The Problem in This Example

When we **toggle the selector** using `setState`, the `BlocSelector` widget does **not re-evaluate the selector** immediately. As a result:

- Even though we’ve updated the selector logic in `setState`, the widget still shows stale data.
- The new selector is only respected **after a new event is emitted** (e.g., tapping Increment).
- This leads to unexpected behavior where the UI **doesn’t reflect the correct selected state**, despite the selector having changed.

---

#### Expected Behavior

When the selector function passed to BlocSelector changes — for example, through a `setState` call in a StatefulWidget — the widget should recompute the selected value using the new selector, even if the bloc’s state hasn’t changed.

The widget should rebuild and reflect the new selected state immediately, based on the updated logic, without requiring a new event to be emitted from the bloc.

### ✅ Workaround

Wrap the `BlocSelector` with a `ValueKey(selected)` to force it to rebuild when the selected tab changes:

```dart
BlocSelector<..., ..., ...>(
  key: ValueKey(selected), // ← Needed for correct behavior
  selector: ...,
  builder: ...,
)
```
Hence

A temporary fix is to **wrap the `BlocSelector` with a `ValueKey`** that changes when your selector logic changes. This forces the widget to rebuild and re-evaluate the new selector.

```dart
BlocSelector<CounterCubit, int, int>(
  key: ValueKey(useEvenSelector), // 👈 forces rebuild on selector toggle
  selector: (state) =>
      useEvenSelector ? evenSelector(state) : oddSelector(state),
  builder: (context, selected) {
    return Text("Selected: $selected");
  },
)
```

> 💡 While this works, it feels more like a workaround than an intended behavior, and rebuilding via keys shouldn’t be necessary just to respect selector updates.

nvshah:
Ensure BlocSelector rebuilds correctly when selector function is updated. Closes #4414

---

## Status

**READY**

---

## Breaking Changes

**NO**

---

## Description

This PR addresses an issue in the BlocSelector widget where a rebuild was not triggered when the selector function itself changed. **_The fix ensures that when the selector function is updated, the widget will rebuild to reflect the new selector logic. This change improves the behavior of the BlocSelector widget, making it more reliable when used with dynamic selector functions_**.

---

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel (approved):
LGTM thanks so much for the fix! 💙

PR #4520, opened 2025-06-05
add flag to catch states emitted from constructor

ktul15:
## Status

**READY**

## Breaking Changes

NO

## Description

I have added a flag "allowInitialState" in blocTest function to catch the states emitted by constructor. When this flag is true, it adds the initial state of bloc/cubit into the "states" array before starting to listen the stream.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel (commented):
I don't think this is the right change. In general I don't recommend emitting from the constructor for this any many other reasons but if we are to address this issue, I think we would need to adjust the timing.

Closing this for now since I don't think this is the right solution but I appreciate you taking the time to contribute!

PR #4536, opened 2025-06-09
docs: fixed docs for custom storage

tmaihoff:
## Status

**READY**

## Breaking Changes

NO

## Description

Just fixed the docs to reflect the actual behaviour.
the `storage` super parameter is not named.
```dart
abstract class HydratedCubit<State> extends Cubit<State>
    with HydratedMixin<State> {
  /// {@macro hydrated_cubit}
  HydratedCubit(State state, [Storage? storage]) : super(state) {
    hydrate(storage: storage);
  }
}
```
## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [x] 📝 Documentation
- [ ] 🗑️ Chore

PR #4537, opened 2025-06-11
fix(hydrated_bloc)!: `HydratedCubit` storage override should be a named parameter

felangel:
## Status

**READY**

## Breaking Changes

YES

## Description

- fix(hydrated_bloc)!: `HydratedCubit` storage override should be a named parameter (#4536)
  - This slipped through the cracks unfortunately 😞 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
> thanks for fixing! lgtm

Thanks for catching this! I'm planning to first land #4390 and do a minor bump and then will land this as part of a major version bump.

gautam-dev24:
@felangel  hi just a small question, don't if it's the right place to ask,  how does the hydrated cubit storage works? 

Suppose i have two different instance of the cubit so they both will have different storage , so how does it get's it's data upon restarts? when app kills and resumes, we are creating new instances of everything right?

this might sound like a silly question

felangel:
> @felangel hi just a small question, don't if it's the right place to ask, how does the hydrated cubit storage works?
> 
> Suppose i have two different instance of the cubit so they both will have different storage , so how does it get's it's data upon restarts? when app kills and resumes, we are creating new instances of everything right?
> 
> this might sound like a silly question

By default all instances use the same storage that you’ve initialized before calling runApp. You can however also override the storage on an instance by instance basis if you need.

gautam-dev24:
@felangel  i'm using get_it and registered my cubit as a factory. using it in two different screens Provided to them at Screen Level. i have a _count = 0 variable in cubit and it's hydrated.

when i changed _count to 2, it was saving fine. but when i navigated to that another screen, a new cubit instance got created as expected only but now _count became 0 again.

it should’ve loaded 2 from storage via fromJson, right? i was assiging in the fromJson also, since storage is shared. to fix this, i made the cubit global instead of per screen.

 what i was doing wrong here?

felangel:
> @felangel i'm using get_it and registered my cubit as a factory. using it in two different screens Provided to them at Screen Level. i have a _count = 0 variable in cubit and it's hydrated.
> 
> when i changed _count to 2, it was saving fine. but when i navigated to that another screen, a new cubit instance got created as expected only but now _count became 0 again.
> 
> it should’ve loaded 2 from storage via fromJson, right? i was assiging in the fromJson also, since storage is shared. to fix this, i made the cubit global instead of per screen.
> 
> what i was doing wrong here?

Are you sure the new instance was actually created after you navigated to the new screen? If you're having trouble I'm happy to take a look at a minimal reproduction sample 👍

gautam-dev24:
I tried re - creating it  , but in the minimal version it's working as expected all the cubits are properly getting iniatlised with the value, even if the instances are different, maybe something else was conflicting as it was a very big codebase and shared.

Thank you for this.

i also have a different doubt, in isolates, i wanted to trigger native method when i receive a firebase push notification,, 

in foreground it's working fine i am able to interact with the native channels... 

but in background or when the app is in killed mode, when the firebase notification is coming and if i am trying to access this method it's throwing unimplemented error,  this method channel code in dart is in a service class which i am calling the firebase _onbackground message received handler.

so how can we achieve this? to trigger any piece of code when our app is in complete kill mode, also interactivity with native channels

PR #4548, opened 2025-06-30
docs(bloc_concepts): update zh-cn copy

lm83680:
Correct improper word usage


felangel (approved):
Thanks for the contribution! 💙

PR #4552, opened 2025-07-02
make color diff optional

### LINKED ISSUE #3465: feat: bloc_test: toPrettyString optional
**Description**

We are runinng tests in CI environment and trying to convert flutter test results into JUnit.xml format in order to display test results in CI environment (Azure DevOPS, but in Gitlab for instance it is similar). 

However Tests suite JUnit.xml format using XML version 1.0 which forbids to use some special characters such as escape characters for colors. 

Would be possible to configure if we want colored outpot from `blocTest` ? 

This ouput of diff is problematic
```
==== diff ========================================

&#x1B;[90m[SplashState.initial(), SplashState.noProfiles()&#x1B;[0m&#x1B;[31m[-, SplashState.noProfiles()-]&#x1B;[0m&#x1B;[90m]&#x1B;[0m

==== end diff ====================================
````

mx1up:
blocTest uses escape codes to color the diff output. This PR makes the color output optional. 

## Status

**READY**

## Breaking Changes

NO

## Description

`blocTest` uses escape codes to color the diff output. This can create problems, see https://github.com/TOPdesk/dart-junitreport/issues/43 and https://github.com/felangel/bloc/issues/3465. 

This PR makes the color output optional through an env var `BLOC_TEST_COLOR_OUTPUT`. (naming up for discussion)

fixes #3465 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mx1up:
if you are okay with this change, i will add it to documentation

mx1up:
@tenhobi i updated the PR title to fix the pipeline

mx1up:
hi @felangel , thanks for your feedback. However, I don't believe checking if the environment supports ansi escapes is enough, see my comment on the issue:

> Checking if the terminal supports ansi escape sequences is not enough, since we want to pass the output onto tojunit to make xml, see https://github.com/TOPdesk/dart-junitreport/issues/43 for more info. For now, a workaround it to sanitize the test output before passing it to tojunit.

https://github.com/felangel/bloc/issues/3465#issuecomment-3028301764

This is basically because the terminal is not the end consumer, but the `tojunit` util. The output has to be limited to xml 1.0, so we cannot include the escape characters in the xml. Even if the current environment supports these escape characters, we still don't want them in the output.

Additionally, the `supportsAnsiEscapes` does not seem to work reliably: I tested it in a terminal supporting escape codes, but the method still returns `false`. The documentation mentions it checks the presence of `xterm` in the `TERM` env var, but that does not seem to be true as I have `TERM=xterm-256color`.

All that being said and thinking more about it, I don't think it's up to bloc_test to provide this option. The generated test json output containing escape characters is valid json. Also, there's no way to receive flutter's `--no-color` option. Other packages might output color as well. So now I think it is either up to `tojunit` to filter its input (not realistic since the package is dead), or filter it in the shell before passing it to `tojunit` (current workaround).

If you agree with my reasoning, feel free to close the PR. Otherwise, I'm happy to improve it.

The only thing `bloc_test` could do, is mention this workaround in the readme/faq.

felangel:
Based on your recent comment I'm inclined to close this PR but I welcome any PRs to update the README/docs, thanks!

PR #4557, opened 2025-07-07
Add Intentions for cubits

c4rl2s0n:
## Status

**READY/IN DEVELOPMENT**
I was not able to deploy it on my machine, but I wanted to add support for creating BlocBuilder etc for working with Cubits, as I got annoyed by always changing Bloc for Cubit when using these templates.

## Breaking Changes

NO

## Description
This PR just adds Intentions to the Intellij plugin that allows to create BlocBuilder etc. with the Cubit-Suffix instead of the Bloc-Suffix

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
Thanks for the PR! I'll take a closer look shortly. I want to see if there is a better way to address the problem you're facing without duplicating all the templates.

c4rl2s0n:
Thank you! That would be great! I just tried to take what I saw and adapt it for my needs. If it would be possible to put like a flag in the template or put a switch in the settings of the plugin, that might also be great.
I am not sure if showing both options (BLoC and Cubit) in the context menu would bloat it unnecessarily, so defining in the settings if one would prefer the intentions for bloc or cubit might be sufficient (not sure how commonly blocs and cubits are mixed, I mostly use cubits for now)

Issue #4456, opened 2025-05-04
feat(bloc_lint): prefer_context_extensions


felangel:
**Description**

As a developer, I want to be warned if I am not using context extensions (e.g. `context.read`, `context.watch`) (originally referenced in #4264)

PR #4562, opened 2025-07-11
feat(bloc_lint): add `prefer_build_context_extensions`

Eittipat:
## Status

**READY**

## Breaking Changes

NO

## Description

feat(bloc_lint): prefer_context_extensions #4456

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/docs/src/content/docs/lint-rules/prefer_build_context_extensions.mdx
+++ b/docs/src/content/docs/lint-rules/prefer_build_context_extensions.mdx
@@ -29,6 +29,12 @@ For consistency, prefer using `BuildContext` extensions like `context.read`,
 `context.watch`, and `context.select` instead of `BlocProvider.of`,
 `RepositoryProvider.of`, `BlocBuilder` or `BlocSelector`.
 
+| explicit method                                                     | extension             |
+| ------------------------------------------------------------------- | --------------------- |
+| `BlocProvider.of<Bloc>(context, listen: false)`                     | `context.read<Bloc>`  |
+| `BlocBuilder<Bloc, State>(...)` or `BlocProvider.of<Bloc>(context)` | `context.watch<Bloc>` |
+| `BlocSelector<Bloc, State>(...)`                                    | `context.select`      |
+
 ## Examples
 
 **Avoid** using `BlocProvider.of<T>(context)` to access a `Bloc` instance.


Eittipat:
Hi @felangel 
I've added some lint rules and hope they align with the requirements.
Please Let me know if there’s anything I can improve.

packages/bloc_lint/lib/src/rules/prefer_context_extensions.dart
```diff
@@ -0,0 +1,43 @@
...
+
+  final LintContext context;
+
+  @override
+  void handleIdentifier(Token token, IdentifierContext _) {
+    if (token.lexeme == 'of') {
+      final prev = token.previous;
+      if (prev != null && prev.lexeme == '.') {
+        final target = prev.previous;
+        if (target != null && target.lexeme == 'BlocProvider') {
```
felangel, line 33: I don't think this is quite right because there are also `RepositoryProvider` extensions that this rule should apply to. In addition, I think we should try to unify some of the logic across this PR and #4572 since they are closely related.

Eittipat: I've updated the logic to support `RepositoryProvider` as well. I also reviewed https://github.com/felangel/bloc/pull/4572. In my case, detecting `RepositoryProvider` and `BlocProvider` usage is much simpler when handled separately. Combining the two makes the logic harder to read and understand. So, I believe it's not necessary to unify them, since they serve different purposes.

felangel: Sounds good will review shortly, thanks for your contributions! 💙

felangel:
I'll take care of resolving the conflicts and getting this merged shortly, thanks!

felangel (changes requested):
Thanks for all your contributions! I really appreciate it 🙏 
Unfortunately, I think this PR still needs some work because it currently only accounts for `BlocProvider.of` and `RepositoryProvider.of` but there are many other cases that aren't accounted for in these changes. For example, I would expect this lint rule to lint when using `BlocBuilder` and recommend using `context.watch` instead. I would also expect this lint rule to lint when using `BlocSelector` and recommend using `context.select` instead.

Let me know if you want to rework this PR to incorporate those changes or if you'd prefer me to take it over and thanks again for taking the time to contribute 💙

Eittipat:
@felangel  I will take care of it. However, I think I need a clearer requirement—like, if `BlocBuilder` has `buildWhen`, should `context.watch` still be recommended?"

felangel:
> @felangel I will take care of it. However, I think I need a clearer requirement—like, if `BlocBuilder` has `buildWhen`, should `context.watch` still be recommended?"

Thanks! For now I’d always recommend context.watch for BlocBuilder (eventually I want to deprecate and remove buildWhen).

felangel:
> @felangel I will take care of it. However, I think I need a clearer requirement—like, if `BlocBuilder` has `buildWhen`, should `context.watch` still be recommended?"

@Eittipat have you had a chance to work on this? If not that’s totally fine I’m happy to help finish this off.

Eittipat:
@felangel I was planning to work on this over the weekend, but I ended up being a bit busy. Feel free to take it.

Thank you

Eittipat:
@felangel Any update on this

felangel:
> @felangel Any update on this

Planning to pick this up now, sorry for the delay!

felangel (approved):
LGTM thanks so much for the contribution! 💙

PR #4572, opened 2025-07-20
feat(bloc_lint): add avoid_context_extensions lint rule

ricardodalarme:
## Status

**READY/IN DEVELOPMENT/HOLD**

## Breaking Changes

YES | NO

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/bloc_lint/lib/src/rules/avoid_context_extensions.dart
```diff
@@ -0,0 +1,49 @@
...
+  @override
+  Listener? create(LintContext context) {
+    return _Listener(context);
+  }
+}
+
+class _Listener extends Listener {
+  _Listener(this.context);
+
+  static const _contextExtensions = [
```
felangel, line 23: I don’t think this is quite right because someone can be using pkg:provider and should still be able to use context extensions on non-bloc objects without any lints

ricardodalarme: Yeap, you are right. I've just updated the implementation to address your point, and it now only flags context extensions on types whose names end with the `Bloc` or `Cubit` suffix, which avoids incorrect lints on other packages like provider. I had to rely on this naming convention because, with the current parser listener architecture, it isn't feasible to inspect the full class hierarchy to programmatically verify if a type extends `Bloc` or extends `Cubit`. Let me know what you think

packages/bloc_lint/lib/src/rules/avoid_context_extensions.dart
```diff
@@ -0,0 +1,98 @@
...
+      final typeToken = openBracketToken.next;
+      if (typeToken != null &&
+          typeToken.type == TokenType.IDENTIFIER &&
+          _isBlocTypeName(typeToken.lexeme)) {
+        _report(token);
+        return;
+      }
+    }
+
+    // Case 2: Inferred type -> final MyBloc bloc = context.read();
```
felangel, line 80: I don't think this is a valid case because `context.read` with an implicit dynamic will fail the lookup 🤷

Issue #1829, opened 2020-10-14
[Test] Question: is blocTest always emit first state even first state equals initial state?


Peng-Qian:
#### Question
Is blocTest will always emit the state even the first state equals the initial state?

I notice there is [a breaking change](https://github.com/felangel/bloc/issues/1512), is this a side-effect of it?

```dart
    blocTest<SomeBloc, SomeState>(
      'when the first state is equal to initial state',
      build: () => SomeBloc(),
      act: (bloc) async {
        // the event just do: yield state
        bloc.add(const SomeEvent.doSomething());
      },
      // in the previous version, should be expect:[] with skip: 1,
      expect: [
        SomeState.initial(),
      ],
      skip: 0,
    );
```

felangel:
Hi @Peng-Qian 👋 
Thanks for opening an issue!

That was an intentional change made to allow blocs to "reignite" (re-notify) listeners if needed when they are created. This is working as expected and if you'd like you can just skip it 👍 

Hope that helps!

mzdm:
Oh, that's why! I spent too much time finding out why when testing Bottom Navigation Bar when the initial index is 0. When I changed to the same index, [0] was emitted (I thought it should be empty). But when doing it after switching to 1st index and then twice times 0 index - then expected was [1, 0] - so 0 index was emitted only once. Glad I found this 😃

PR #4630, opened 2025-10-03
fix(hydrated_bloc): prevent duplicate emissions after hydration

### LINKED ISSUE #4549: fix: HydratedMixin causes BlocBase to emit the same state
**Description**
When a hydrated bloc gets restored to state X, if you try to emit the same state X again, it will actually get emitted. This seems like an antipattern, because developers are used to not having to check if the state is different before emitting: 

```dart
if(state != newState) {
  emit(newState)
}
```

This is also a fairly common use case, which makes it cumbersome to deal with. For example: 
1. you hydrate a bloc that fetches data over network, because it makes the subsequent app startups faster and the data rarely changes.
2. When opening the app, you optimistically assume the data hasnt changed, but issue a network request in the background just to validate that assumption
3. If the request comes back with the same data, there is no need to emit a state that would trigger downstream listeners, builders, etc. But at the moment it does, so developers need to remember to add guard statements specifically for this, rather than relying on the built-in equality operator. From the docs: https://pub.dev/documentation/bloc/latest/bloc/BlocBase/emit.html

```
emit does nothing if the state being emitted is equal to the current state.
```

**Steps To Reproduce**
<details>

<summary>Minimum reproducible example</summary>

```dart
import 'package:hydrated_bloc/hydrated_bloc.dart';

class CounterCubit extends Cubit<int> with HydratedMixin {
  CounterCubit() : super(0) {
    hydrate();
  }

  void setToFive() {
    emit(5);
  }

  @override
  int? fromJson(Map<String, dynamic> json) {
    return 5;
  }

  @override
  Map<String, dynamic>? toJson(int state) {
    return {'state': 5};
  }
}

class MyHydratedStorage implements Storage {
  @override
  dynamic read(String key) {}

  @override
  Future<void> write(String key, dynamic value) async {}

  @override
  Future<void> delete(String key) async {}

  @override
  Future<void> clear() async {}

  @override
  Future<void> close() async {}
}

void main() {
  HydratedBloc.storage = MyHydratedStorage();
  CounterCubit()
    ..stream.listen(print)
    ..setToFive();
}
```

</details>

**Expected Behavior**
Nothing should get printed, as the `setToFive` method tries to emit the same state that the cubit is already in (as a result of hydration)

**Screenshots**

**Additional Context**
I assume this happens because of [this](https://github.com/felangel/bloc/blob/f4c917fd861fcec7cec06e2288d9144b92b4d868/packages/bloc/lib/src/bloc_base.dart#L102) `_emitted` check, which doesnt get set to true in the `hydrate` method. Is this by design? I cant find any documentation around it if yes.

realmeylisdev:
## Summary
Fixes #4549 

When a HydratedBloc/HydratedCubit is restored from storage, emitting the same state now correctly does nothing instead of emitting it to the stream.

## Problem
The issue occurred because the `_emitted` flag in `BlocBase` was not being set during hydration. This meant that the first `emit()` call after hydration would bypass the equality check (since `_emitted` was still `false`), causing duplicate state emissions.

## Solution
- Modified `hydrate()` to use `emit()` when restoring state, which properly sets the `_emitted` flag in `BlocBase`
- Added `_hydrationInProgress` flag to prevent `onChange` from writing to storage during the hydration process (avoiding conflicts)
- Added regression test (`issue_4549_test.dart`) to verify the fix

## Test plan
- [x] All existing hydrated_bloc tests pass (73 tests)
- [x] New regression test validates the fix
- [x] Code formatted with `dart format`
- [x] Code analyzed with `dart analyze --fatal-infos --fatal-warnings` - no issues

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Includes test coverage

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes


felangel:
Thanks for the PR! I need to think about this a bit more carefully because this will potentially break any apps which depend on `BlocListeners` being triggered by state hydration.

realmeylisdev:
Hey @felangel, would you mind reviewing this?

felangel:
> Hey @felangel, would you mind reviewing this?

This is on my todo list for this week. Apologies for the delay!

felangel:
@realmeylisdev I've thought about this some more and I actually think this is working as expected. If you think of hydrated_bloc as just automatically persisting/restoring the state then a HydratedBloc with a hydrated state of 42 should behave the exact same as a Bloc with an initial state of 42. In both cases, emitting 42 should trigger a change event and emit the state in order to be consistent and support https://github.com/felangel/bloc/issues/1829#issuecomment-708435624

In other words, we should not think of a "hydration" as equivalent to emitting a new state. It should be thought of as an adjustment to the initial state instead.

Let me know what you think, thanks!

felangel:
Going to close for now per my above comment but I'm more than happy to continue the conversation 👍

Issue #4264, opened 2024-10-10
feat: Custom lint rules


zbarbuto:
It might be cool to add some custom lint rules to enforce consistency and/or good practices when using Bloc. 

Some random suggestions and ideas (I haven't thought too long on these, some might be outright wrong, just starting discussion):
- State classes should end with `State` suffix
- Enforce immutable state classes
- Ensure state/event classes are `sealed`
- Prefer `context.select()` over `BlocBuilder` unless providing a `buildWhen` argument
- Avoid adding public methods on Bloc classes (i.e. not cubits)
- Avoid public properties on Blocs and Cubits besides `state`
- Avoid `BuildContext`s as arguments to cubit methods / bloc events
- Prefer `context.read<T>()` over `BlocProvider.of<T>(context)` (or vice/versa) for consistency
- Ensure the event superclass or all its subclasses are handled by `on<T>`s in constructor
- Avoid using the result of a cubit call (to prefer listening for state changes instead)
- Detect accidentally providing a bloc with `Provider<MyBloc>` or `RepositoryProvider<MyBloc>` instead of `BlocProvider<MyBloc>`
- Avoid passing a bloc or cubit to the constructor of another bloc or cubit

Rules could be opt-in but it would also be good to have a "recommended" set which are accepted as general best-practices.

**Related work:**
- [riverpod](https://pub.dev/packages/riverpod_lint) has some pretty extensive lint rules and custom fixers to go along with them.
- The Angular NgRx state management package's [ESLint rules](https://ngrx.io/guide/eslint-plugin) which catch common mistakes
- [DCM](https://dcm.dev/) offers several [bloc-centric lints](https://dcm.dev/docs/rules/?categories=bloc) under their paid plan

tenhobi:
Hi, a bit off topic here: some of these lints/rules are already in the DCM, which I highly recommend. https://dcm.dev

zbarbuto:
I did come across DCM shortly after posting this. Good for inspiration, but my preference would still be if the lints were first-party to the library itself for various reasons as well as open source to allow contributions.

Also, all the bloc-based lints appear to be behind the Pro+ plan, which wouldn't be for everyone.

sanba-anass:
> ich I highly recommend. https://dcm.dev
nice try

vmichalak:
Because i really like this idea, i've start a project named [bloc_lint](https://github.com/vmichalak/bloc_lint) to purpose something.
Feel free to contribute 🤗

[...]

PR #4665, opened 2025-11-11
feat(bloc_lint): add `avoid_bloc_to_bloc_members`

zbarbuto:
## Status

**READY**

## Breaking Changes

NO

> Maybe? If you use all lint rules this might break if you have all lint rules enable for your projevt?

## Description

Adds support for a new `avoid_bloc_to_bloc_members` lint rule (I'm not married to the name, open to suggestions).

This was mentioned in #4264 as "Avoid passing a bloc or cubit to the constructor of another bloc or cubit" (see [this comment](https://github.com/felangel/bloc/issues/4264#issuecomment-2547451118))

The goal being to avoid tightly coupling blocs or cubits together by allowing one to be a member or parameter of another.

❌ Bad

```dart
class TightlyCoupledBloc extends Bloc {
  late final StreamSubscription otherBlocSubscription;

  TightlyCoupledBloc(OtherBloc otherBloc) {
  //                 ^^^^^^^^^^^^^^^^^^^
    otherBlocSubscription = otherBloc.stream.listen((state) {
      add(MyEvent());
    });
  }

  @override
  Future<void> close() {
    otherBlocSubscription.cancel();
    return super.close();
  }
}
```

❌ Bad

```dart
class FooCubit extends Cubit {
   final BarCubit otherCubit;
   //    ^^^^^^^^^^^^^^^^^^^
   FooCubit({ required this.otherCubit }): super();
}
```

Currently this covers both in one rule (as class fields and as parameters) but happy to split into two separate rules if that's preferable - I just figured they both fall under the same umbrella of "avoid tightly coupling blocs or cubits together".

It's done via name rather than looking at type hierarchy as that's what the other rules were doing (e.g. `avoid_public_bloc_methods`) - I'm not very familiar with the analyzer/diagnostic tools so unsure if there's a way to do proper checking on the parameter typings to see if a parameter actually extends `Bloc` or `Cubit`?

Also happy to extend test coverage - I don't have **every single case** for both bloc and cubit and every combination between but tried to cover at least one of every variation (optional vs required, named vs positional vs default, bloc vs cubit).

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
Thanks I'll take a closer look shortly!

zbarbuto:
Thanks - have fixed the formatting issue.
Other CI failure looks like I've missed docs somewhere, but I'm not sure?

PR #4712, opened 2025-12-25
docs(lint-rules): add fa translations

adultcode:
Translated the lint rules pages into Persian to make the docs more accessible for Persian-speaking users. Preserved examples and page structure so behavior and formatting remain unchanged.


felangel (approved):
Looks good to me! Thanks so much for the continued contributions! 💙

PR #4758, opened 2026-03-11
fix(hydrated_bloc): serialize non-string map keys

### LINKED ISSUE #3983: fix: `Enum` with `int` type cannot be stored
**Description**
When I used `@JsonValue` to specify `int` value to an enum entry, the value cannot be stored and the retrieved value will always be empty.

**Steps To Reproduce**

1. Annotate enum values with `@JsonValue` of type `int`
```dart
enum Step{
  @JsonValue(1)
  step1,
  @JsonValue(2)
  step2,
  @JsonValue(3)
  step3,
}
```

2. Create a param in `State` with type `Map<Step, String>`
```dart
@freezed
class RegisterState with _$RegisterState {
  factory RegisterState ({
    @Default({}) Map<Step, String> registerSteps,
  }) = _RegisterState;

  factory RegisterState .fromJson(Map<String, dynamic> json) => _$RegisterStateFromJson(json);
}
```
3. Modify `State`
```dart
emit(state.copyWith(
        registerSteps: {
          Step.step1: 'done',
          Step.step2: 'doing',
          Step.step3: 'none',
        },
)
```
4. Hot Restart to check state

**Expected Behavior**

The state should be 
```dart
registerSteps: {
   Step.step1: 'done',
   Step.step2: 'doing',
   Step.step3: 'none',
},        
```

**Actual Behavior**

The state showed
```dart
registerSteps: {},        
```

**Additional Context**

The problem is in the method `dynamic _traverseComplexJson(dynamic object)`,

```dart
    } else if (object is Map) {
      _checkCycle(object);
      final map = <String, dynamic>{};
      object.forEach((dynamic key, dynamic value) {
        final castKey = _cast<String>(key);
        if (castKey != null) {
          map[castKey] = _traverseWrite(value).value;
        }
      });
      _removeSeen(object);
      return map;
    }
```

The map's key is type  `int` so it cannot be casted to `String` so the value will not be stored

ersanKolay:
## Summary

- Fixes `_traverseComplexJson` silently dropping map entries with non-String keys (e.g. int keys from `@JsonValue(int)` enums used as map keys in `json_serializable`).
- The root cause was `_cast<String>(key)` returning `null` for non-String keys, causing entries to be skipped entirely. Replaced with `key is String ? key : key?.toString()`, consistent with how `_traverseRead` already handles keys.
- Added unit tests and E2E tests covering non-string map key serialization and restoration.

Closes #3983

## Test plan

- [x] Existing 122 tests pass
- [x] New unit test: verifies int map keys are converted to string keys when persisting
- [x] New unit test: verifies state restoration from cache with stringified int keys
- [x] New E2E test: full write-read-restore cycle with `@JsonValue(int)` enum-like map keys
- [x] `dart analyze` clean (no errors or warnings)

### DIFF
--- a/packages/hydrated_bloc/test/cubits/season_palette_cubit.dart
+++ b/packages/hydrated_bloc/test/cubits/season_palette_cubit.dart
@@ -22,9 +22,9 @@ class SeasonPalette {
   const SeasonPalette(this.colors);
 
   factory SeasonPalette.fromJson(Map<String, dynamic> json) {
-    final deserialized = json['colors'] as Map<String, dynamic>? ?? {};
+    final raw = json['colors'] as Map<String, dynamic>? ?? {};
     return SeasonPalette(
-      deserialized.map(
+      raw.map(
         (key, value) => MapEntry(
           Season.fromJson(int.parse(key)),
           value as String,


felangel (dismissed):
LGTM thanks so much for the contribution! 💙

PR #4759, opened 2026-03-11
fix(bloc): make isClosed return true immediately when close is called

### LINKED ISSUE #4749: isClosed is false although the bloc is closed
**Description**
I have a page with its own bloc instance. The bloc has the following code:
```dart
  Future<void> _onSubscriptionRequested(
    EntriesSubscriptionRequested event,
    Emitter<EntriesState> emit,
  ) async {
    emit(state.copyWith(status: EntriesStatus.loading));
    try {
      await emit.onEach<EntriesPageData>(
        _watchEntriesUsecase.call(event.fieldListId),
        onData: (entriesPageData) {
          if (!isClosed) {
            add(NewData(entriesPageData));
            add(PrepareTab(state.currentTabIndex, DateTime.now()));
          }
        },
        onError: (_, _) => emit(state.copyWith(status: EntriesStatus.failure)),
      );
    } catch (e) {
      emit(state.copyWith(status: EntriesStatus.failure));
    }
  }
```
When I click the FloatingActionButton, a new page is opened. The new page adds new data, and the onData of the previous page is called as the stream emitted the new data. But `Bad state: Cannot add new events after calling close` is thrown, although the ` if (!isClosed)` check!

**Expected Behavior**
`Bad state: Cannot add new events after calling close`  should not be thrown. Altought this would mean that the event for new data would not be added!

**Additional Context**
I add a new event instead of updating the state directly or using methods, because I need to test my bloc when new data is emitted by the stream after the first one. In the todo example, I find how to test the first emitted data, but no tests for the second and the next emitted data. In the tests, I use the act function to add an event and test the expected state.

ersanKolay:
## Summary

- Fixes a race condition where `isClosed` returned `false` during `Bloc.close()` teardown, but `add()` would throw a `StateError` because the internal `_eventController` was already closed.
- Root cause: `isClosed` was based on `_stateController.isClosed`, but `Bloc.close()` closes `_eventController` first. This created a window where checking `isClosed` before `add()` was unreliable.
- Introduces a `_closed` flag in `BlocBase` set at the start of `close()`, and overrides `isClosed` in `Bloc` to also check `_eventController.isClosed`.
- Internal `emit`/`onEmit` checks use `_stateController.isClosed` directly so in-progress event handlers can still emit states during teardown.

Closes #4749

## Test plan

- [x] All 118 existing tests pass
- [x] New test: `Bloc.isClosed` returns true immediately when `close()` is called (without await)
- [x] New test: `Cubit.isClosed` returns true immediately when `close()` is called (without await)
- [x] `dart analyze` clean (no errors or warnings)

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -126,6 +126,9 @@ abstract class Bloc<Event, State> extends BlocBase<State>
     _blocObserver.onEvent(this, event);
   }
 
+  @override
+  bool get isClosed => _closed || _eventController.isClosed;
+
   /// {@template emit}
   /// **[emit] is only for internal use and should never be called directly
   /// outside of tests. The [Emitter] instance provided to each [EventHandler]
@@ -196,7 +199,7 @@ abstract class Bloc<Event, State> extends BlocBase<State>
       _eventController.stream.where((event) => event is E).cast<E>(),
       (dynamic event) {
         void onEmit(State state) {
-          if (isClosed) return;
+          if (_stateController.isClosed) return;
           if (this.state == state && _emitted) return;
           onTransition(
             Transition(
--- a/packages/bloc/lib/src/bloc_base.dart
+++ b/packages/bloc/lib/src/bloc_base.dart
@@ -69,6 +69,8 @@ abstract class BlocBase<State>
 
   bool _emitted = false;
 
+  bool _closed = false;
+
   @override
   State get state => _state;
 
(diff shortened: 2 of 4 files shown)


felangel:
Thanks for the PR but I’m not sure this is the correct fix. I think the first step here is to make a minimal reproduction sample. The tests you added don’t reflect the issue linked.

ersanKolay:
Thanks for the feedback. I added a minimal reproduction sample as a comment on #4749 and updated the PR with a test that reproduces the exact scenario from the issue: a Bloc using `emit.onEach` where `onData` guards `add()` with `isClosed`, but external stream data during `close()` teardown causes a `StateError`.

Let me know if you'd like a different approach for the fix itself.

felangel:
> Thanks for the feedback. I added a minimal reproduction sample as a comment on #4749 and updated the PR with a test that reproduces the exact scenario from the issue: a Bloc using `emit.onEach` where `onData` guards `add()` with `isClosed`, but external stream data during `close()` teardown causes a `StateError`.
> 
> Let me know if you'd like a different approach for the fix itself.

Thanks I'll take a look shortly! 🙏

felangel:
Thanks for the PR however I'm going to close in favor of https://github.com/felangel/bloc/pull/4809 which is the proper fix imo.

PR #4792, opened 2026-04-15
feat: add "Don't Show Again" option in the VSCode extension

### LINKED ISSUE #4726: feat: make it easier to turn off the version notifications
**Description**

The version notifications on startup are a bit annoying, tbh. I think in general such notifications should be kept down to a minimum.

**Desired Solution**

Maybe a "Don't show again" could be added, so it's easier to disable, rather than having to go into the settings or blocking notifications for the whole plugin.

haashem:
## Status

**READY**

## Breaking Changes

NO

## Description

Add `Don't Show Again` option in the VSCode update notification, to suppress update notification which after a while becomes annoying.

Resolves #4726  

dontshow

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/extensions/vscode/src/utils/analyze-dependencies.ts
+++ b/extensions/vscode/src/utils/analyze-dependencies.ts
@@ -1,6 +1,6 @@
 import * as _ from "lodash";
 import * as semver from "semver";
-import { window, env, Uri } from "vscode";
+import { window, env, Uri, workspace } from "vscode";
 import { getLatestPackageVersion, getPubspecLock } from ".";
 import { updatePubspecDependency } from "./update-pubspec-dependency";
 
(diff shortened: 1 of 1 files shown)


felangel (approved):
LGTM thanks for the contribution 🙏

felangel:
I've thought about it a bit and I don't think we should do this because it introduces a new source of confusion:

> if I click "don't show again", then how to I undo that? 

See https://github.com/microsoft/vscode/issues/23314 for an example of this confusion in a different context.

I also think it's already very simple to open the extension settings and uncheck a box so this doesn't really feel necessary imo. Happy to be convinced otherwise if you feel strongly. Thank! 🙏

haashem:
Thanks for reviewing this and for the valuable point. Based on my years of using this extension, I never realized I could disable this constant popup from the extension settings 😳. I only discovered it a few weeks ago after finding the related issue and your reply.

I think most users simply tolerate the slight annoyance of clicking the “OK” button every day and immediately jumping back to their workspace, rather than looking for a way to disable the popup.

Another idea proposed in the linked issue is to show a small hint message to users, so they know the popup can be disabled from the extension settings. We can follow the same here. What do you think? for me it improves the UX.

PR #4797, opened 2026-04-24
fix(bloc_tools): add --no-version-check flag

### LINKED ISSUE #4787: fix: bloc_tools' version check and confirmation does not detect non-interactive terminal reliably.
**Description**

**Steps To Reproduce**

bloc_tool's command_runner.dart performs a version check. When it detects that a newer version is available, it asks if it should update. If running in a non-interactive environment, the confirmation prompt needs to detect this; otherwise it would just hang.

Unfortunately this detection doesn't always work right. When I run bloc_tool from OpenCode, it thinks it's running interactively and as a result block.

I don't know why exactly it thinks it's running interactively. But I would suggest at least adding a command line parameter that disables those version checks. Because even if running interactively, where I can just decline, it's still annoying getting this unrequested behavior. Especially as I can't always just update to get rid of the notice, as sometimes version constraints prevent that.

**Expected Behavior**
Ideally just don't do the version check. Alternatively, please provide a command line flag to disable it.

nahyunuk:
## Status
READY

## Breaking Changes
NO

## Description
Fixes #4787

Added `--no-version-check` flag to `bloc_tools` to allow users 
to skip the version check prompt in non-interactive environments 
or when the behavior is undesired.

### Usage
```sh
bloc --no-version-check new bloc my_bloc
```

## Type of Change

### DIFF
--- a/packages/bloc_tools/lib/src/command_runner.dart
+++ b/packages/bloc_tools/lib/src/command_runner.dart
@@ -28,6 +28,11 @@ class BlocToolsCommandRunner extends CommandRunner<int> {
       negatable: false,
       help: 'Print the current version.',
     );
+    argParser.addFlag(
+      'no-version-check',
+      negatable: false,
+      help: 'Skip the version check.',
+    );
     addCommand(
       LanguageServerCommand(languageServerBuilder: languageServerBuilder),
     );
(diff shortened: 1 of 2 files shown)


felangel:
Thanks for the PR! I think a better solution would be configuring an environment variable rather than having to specify a flag on every single command invocation.

Closing for now in favor of https://github.com/felangel/bloc/pull/4798

PR #4814, opened 2026-05-17
refactor(flutter_firebase_login): preserve original stack traces (#2995)

### LINKED ISSUE #2995: refactor: stop losing the original stack trace from dart 2.16
You must stop losing the original stack trace like this:

https://github.com/felangel/bloc/blob/a3a9b2dba6546480d4ea26931ed024f274b9a7a9/examples/flutter_firebase_login/packages/authentication_repository/lib/src/authentication_repository.dart#L206-L210

from dart 2.16 you must use [Error.throwWithStackTrace](https://api.dart.dev/dev/2.16.0-35.0.dev/dart-core/Error/throwWithStackTrace.html) like this:

```dart
//@dart=2.16

...

} on FirebaseAuthException catch (e, stackTrace) {
  Error.throwWithStackTrace(
    SignUpWithEmailAndPasswordFailure.fromCode(e.code),
    stackTrace,
  );
}
```

also, instead:
```dart
} catch (_) {
```
you must write:
```dart
} on Object {
```

sulphur-swarm:
## Summary

Closes #2995

This refactor updates `AuthenticationRepository` in the `flutter_firebase_login` example to preserve the original stack traces when re-throwing exceptions.

## Changes

**File:** `examples/flutter_firebase_login/packages/authentication_repository/lib/src/authentication_repository.dart`

All four async methods (`signUp`, `logInWithGoogle`, `logInWithEmailAndPassword`, `logOut`) previously lost the original stack trace by throwing a new exception object with `throw`. This makes crash reporting tools (Sentry, Crashlytics, etc.) show a truncated/misleading stack trace.

The fix uses `Error.throwWithStackTrace` (available since Dart 2.16, well within the SDK constraint `>=3.10.0`) to associate the original stack trace with the re-thrown domain exception:

```dart
// Before
} on firebase_auth.FirebaseAuthException catch (e) {
  throw SignUpWithEmailAndPasswordFailure.fromCode(e.code);
} catch (_) {
  throw const SignUpWithEmailAndPasswordFailure();
}

// After
} on firebase_auth.FirebaseAuthException catch (e, stackTrace) {
  Error.throwWithStackTrace(
    SignUpWithEmailAndPasswordFailure.fromCode(e.code),
    stackTrace,
  );
} on Object catch (_, stackTrace) {
  Error.throwWithStackTrace(
    const SignUpWithEmailAndPasswordFailure(),
    stackTrace,
  );
}
```

Additionally, bare `catch (_)` clauses are replaced with `on Object catch (_, stackTrace)` which is the idiomatic Dart pattern — it is explicit about catching all objects while still capturing the stack trace.

## Testing

All existing tests pass (`flutter test`). No new tests were required because the existing tests already cover all modified code paths (they verify the correct exception type is thrown, and `Error.throwWithStackTrace` preserves the type).

---

*This fix was contributed with AI assistance by SwarmFix (sulphur-swarm).*

### DIFF
--- a/examples/flutter_firebase_login/packages/authentication_repository/lib/src/authentication_repository.dart
+++ b/examples/flutter_firebase_login/packages/authentication_repository/lib/src/authentication_repository.dart
@@ -202,10 +202,16 @@ class AuthenticationRepository {
         email: email,
         password: password,
       );
-    } on firebase_auth.FirebaseAuthException catch (e) {
-      throw SignUpWithEmailAndPasswordFailure.fromCode(e.code);
-    } catch (_) {
-      throw const SignUpWithEmailAndPasswordFailure();
+    } on firebase_auth.FirebaseAuthException catch (e, stackTrace) {
+      Error.throwWithStackTrace(
+        SignUpWithEmailAndPasswordFailure.fromCode(e.code),
+        stackTrace,
+      );
+    } on Object catch (_, stackTrace) {
+      Error.throwWithStackTrace(
+        const SignUpWithEmailAndPasswordFailure(),
+        stackTrace,
+      );
     }
   }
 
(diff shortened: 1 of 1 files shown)


felangel (approved):
LGTM thanks for the contribution 💙

Issue #3248, opened 2022-02-28
docs: clearly document semantics for different transformers off the same bloc


msackman:
It's potentially a confusing API design to have the ability to specify a concurrency transformer _per_ event handler - initially I felt that it would be better to have one event transformer per bloc. But I note that e.g. https://github.com/VGVentures/bloc_concurrency_demos/blob/85fd3948b4e77b537f82a59a8e1d4e5fdf417262/lib/registration/bloc/registration_bloc_new.dart does make use of multiple different types of event transformer for handlers off the same bloc.

It's not clear (to me) what the semantics are.

https://pub.dev/documentation/bloc_concurrency/latest/bloc_concurrency/sequential.html states:

> Note: there is no event handler overlap and every event is guaranteed to be handled in the order it was received.

I'm not sure that's the case though - if you have several event handlers on the same bloc, looking at the implementation, it's not at all obvious that setting `sequential` on one would prevent a _different_ event handler from starting concurrently. If a bloc has 2 event handlers, and they both set `sequential` it's still not clear to me that they mutually exclude each other (because I think the async map is happening after the filtration on the stream, so I think you're essentially splitting the stream in 2 parts first, before applying any exclusion semantics; that said, I'm really quite new to dart/flutter/bloc so I could be *very* wrong on all of this).

If it were me, I would shift the transformer to be at the bloc level, and not the event handler level, as I think the semantics are much clearer and easier to understand that way. If there really are compelling reasons to have it per handler, then I think the docs should be really explicit about the fact that having one event handler which does its own type-based dispatch is potentially quite different semantics to having two event handlers and getting the bloc to do the type-based dispatch *even* if those two event handlers are using the same transformer, because of the interaction between (a) splitting the stream per type; (b) applying handlers as a mapping over the (sub) stream(s).

(Again, if it were me, I would have the default transformer be sequential, not concurrent. But I'm sure you've reasons for the choices you've made.)

federicodesia:
Hello!

> If a bloc has 2 event handlers, and they both set sequential it's still not clear to me that they mutually exclude each other.

In that case the events are not mutually exclusive, because they are two different events.

Event transformations depend on the expected behavior, so sometimes you only need to apply a transformer to a specific event and not to the others. If you want to apply a transformer to multiple events, you must group the events like in #3143.

champ96k:
[...]

Issue #3143, opened 2022-01-16
feat: Ability to apply a transformer to all events


Amir-P:
**Description**

Recently I started using bloc v8 and bloc_concurrency and I noticed that in this new version there is no way (or at least I couldn't find any way) to apply a transformer to all incoming events. Right now we can do something like this:

```dart
MyBloc(this.useCase) : super(const MyBlocState()) {
  on<Event1>(_event1);
  on<Event2>(_event2, transformer: debounced());
}
```

But what if we wanted to apply restartable transformer to all events?

**Desired Solution**

Ability to define global transformers.

felangel:
Hi @Amir-P 👋 
Thanks for opening an issue!

If you want to apply a transformer to multiple events you must group the events together. In this case, if you want all events to be `restartable` then you could do:

```dart
MyBloc(this.useCase) : super(const MyBlocState()) {
  on<Event>(_onEvent, transformer: restartable());
}

FutureOr<void> _onEvent(Event event, Emitter<State> emit) {
  if (event is Event1) return _onEvent1(event, emit);
  if (event is Event2) return _onEvent2(event, emit);
}
```

You can read more about this [here](https://verygood.ventures/blog/how-to-use-bloc-with-streams-and-concurrency).

Hope that helps 👍

Amir-P:
Thanks for your response @felangel .

That's something I thought about but doesn't we lose the ability to apply transformers on specific events? For example I want to be able to apply debounce on a specific type of events and apply restartable on all events. How this is achievable?

felangel:
> Thanks for your response @felangel .
> 
> That's something I thought about but doesn't we lose the ability to apply transformers on specific events? For example I want to be able to apply debounce on a specific type of events and apply restartable on all events. How this is achievable?

No problem! You can write a transformer that handles applying the correct transformer depending on the event type. See https://github.com/felangel/bloc/blob/10542d868fb0d615e1f7afba9d6a5af39a221bca/packages/bloc/test/blocs/counter/merge_bloc.dart#L6 for an example.

Amir-P:
Thanks @felangel ! It really helped.

PR #4819, opened 2026-05-27
docs: clarify per-handler event transformer semantics (#3248)

champ96k:
## Status
READY
## Breaking Changes
NO
## Description
Each registered event handler operates on its own independent stream pipeline. Different handlers with different transformers do not affect each other's execution.
### Changes
- Add `Per-Handler Event Transformers` section to `bloc-concepts.mdx` explaining each `on<E>()` call creates an independent subscription that filters the shared event stream by type and applies its own transformer, with a Dart example showing two handlers with different transformers on the same Bloc.
- Clarify doc comments in all `bloc_concurrency` transformers (`sequential`, `restartable`, `droppable`, `concurrent`) that their guarantees apply per-handler, not across handlers.
- Clarify that "no event handler overlap" refers to the specific handler using the transformer, not other handlers registered on the same Bloc.
### Related Issues
Part of #3248
## Type of Change
- [ ] ✨ New feature
- [ ] 🛠️ Bug fix
- [ ] ❌ Breaking change
- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/docs/src/content/docs/bloc-concepts.mdx
+++ b/docs/src/content/docs/bloc-concepts.mdx
@@ -622,5 +622,39 @@ opinionated set of event transformers.
 
 :::
 
+#### Per-Handler Event Transformers
+
+Each event handler registered via `on<E>()` operates on its own independent
+stream pipeline. Internally, every call to `on<E>()` filters the shared event
+stream by the event type `E` and applies the specified transformer to that
+filtered sub-stream. As a result, different handlers with different transformers
+run completely independently — there is no cross-handler coordination.
+
+For example, if a single `Bloc` has two handlers — one with `sequential` and
+another with `restartable` — the two handlers will not affect each other's event
+queues or execution order:
+
+```dart
+class SearchBloc extends Bloc<SearchEvent, SearchState> {
+  SearchBloc() : super(const SearchState()) {
+    // Uses sequential transformer: events of this type are processed
+    // one at a time in FIFO order.
+    on<SearchQueryChanged>(_onQueryChanged, transformer: sequential());
+
+    // Uses restartable transformer: if a new ClearSearch event arrives
+    // before the previous one completes, the previous handler is cancelled.
+    on<ClearSearch>(_onClearSearch, transformer: restartable());
+  }
+}
+```
+
+In the example above, `SearchQueryChanged` events are guaranteed to be handled
+sequentially relative to each other, and `ClearSearch` events use restartable
+semantics — but the two handlers never block or interfere with one another.
+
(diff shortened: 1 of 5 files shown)


GhagSagar23:
Thanks for tackling #3248, @champ96k — the per-handler scoping is exactly the right thing to clarify. Sharing a few suggestions that I think would strengthen the PR, cc @felangel for a maintainer steer:

**1. Use a snippet component instead of a raw code fence.** Every code sample in `bloc-concepts.mdx` is imported as a `<...Snippet />` `.astro` component (`<Code code={...} lang="dart" title="..." />`) rather than an inline ` ```dart ` block — e.g. `DebounceEventTransformerSnippet`. The inline fence in the new section is the only one on the page that breaks that pattern. Extracting it into `docs/src/components/concepts/bloc/…Snippet.astro` keeps it consistent (and picks up the site's syntax highlighting + filename chrome).

**2. Make the example self-consistent.** The current sample references `SearchEvent` / `SearchState` / `_onQueryChanged` / `_onClearSearch` without defining them. A `sealed class SearchEvent` with two `final` subclasses (matching the sealed-class style the rest of the docs now use) makes it copy-paste-coherent.

**3. Answer the follow-up the reporter actually asked.** In #3248, `@msackman` ends on *"how do I keep sequential ordering **across** event types?"* — and `@federicodesia` points at grouping (#3143). The PR explains that handlers are independent but doesn't show the resolution. One extra snippet closes that loop:

```dart
class SearchBloc extends Bloc<SearchEvent, SearchState> {
  SearchBloc() : super(const SearchState.initial()) {
    // A single handler on the shared supertype sequences every subtype.
    on<SearchEvent>(_onEvent, transformer: sequential());
  }

  Future<void> _onEvent(SearchEvent event, Emitter<SearchState> emit) async {
    switch (event) {
      case QueryChanged():
        // ...
      case SearchCleared():
        // ...
    }
  }
}
```

**4. Worth a one-liner on the invariant.** `on<E>` may be registered only once per event type (it throws a `StateError` otherwise) — noting this makes it clear *why* per-handler scoping is the model rather than an accident.

**On the transformer doc comments:** the clarifications read well. Minor: the appended paragraph is the same three lines verbatim across all four transformers — folding the scope note into each existing `**Note**` (and, for `sequential`/`restartable`, pointing at the grouping pattern) keeps them tighter.

Happy to open a PR against your branch with these if that's easier, or you're welcome to lift any of it directly — whatever you and @felangel prefer. Thanks again for pushing this forward. 🙏

PR #4859, opened 2026-08-13
docs: testing initial events

### LINKED ISSUE #4229: docs: testing initial events
**Description**

Hi! 

In #3944, it's mentioned that the recommended way to add initial events to a Bloc is as follows: 

```
BlocProvider(
  create: (context) => MeetingsBloc()..add(LoadMettingsEvent()),
  child: MyWidget(),
);
```

The reasoning being that
> this gives you granular control over when the event is added and makes testing easier and more predictable. In addition, it also makes the bloc more flexible and reusable.

However, this creates a new problem. How do you test that the correct initial events are added when the Bloc is created?
As far as I know, none of the Bloc examples or documentation demonstrate a way to do this.

Is adding the initial events not considered worth testing?

AzazelSensei:
#4229

The first event added in `BlocProvider.create` is widget-tree behavior, not something the bloc owns. I added a short section to the testing docs that shows the existing flutter_todos widget test (pump the page with a mocked repository and verify the subscription).

I left the example test alone. It is the same one called out on the issue and still matches the page.

Closes #4229

### DIFF
--- a/docs/src/content/docs/testing.mdx
+++ b/docs/src/content/docs/testing.mdx
@@ -72,9 +72,51 @@ complex test using the [bloc_test](https://pub.dev/packages/bloc_test) package.
 
 We should be able to run the tests and see that all are passing.
 
+## Testing Initial Events
+
+If you add an initial event in `BlocProvider.create`, that behavior is part of
+the widget tree, not the bloc. Cover it with a widget test instead of a bloc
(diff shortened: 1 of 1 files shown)


AzazelSensei:
Closing. I'm not leaving work parked on a project that doesn't respond.

PR #4862, opened 2026-08-13
feat: add force flag to emit in Bloc and Cubit

OmarAly92:
## Status
READY

## Breaking Changes
NO

## Description
This PR adds a `force` parameter to the `emit()` method across the bloc library, allowing developers to emit duplicate states (states equal to the current state) when needed.

## Changes

- **`BlocBase`**: Added optional `force` parameter to `emit()`. When `force: true`, the state is emitted even if it equals the current state, triggering `onChange` callbacks and notifying listeners.
- **`Bloc`**: Updated the `emit()` override to accept and forward the `force` parameter. Modified the internal `onEmit` callback in event handlers to respect the `force` flag.
- **`Emitter`**: Added `force` parameter to the `call()` method and the `_Emitter` implementation to support forced emissions in event handlers.
- **`ReplayBlocMixin` & `ReplayCubitMixin`**: Updated to forward the `force` parameter through the replay stack operations.
- **Tests**: Added comprehensive test coverage for the new functionality:
  - Cubit tests verify forced emissions trigger `onChange` callbacks and emit duplicate states
  - Bloc tests verify forced emissions work across events and trigger `onTransition` callbacks
  - Tests confirm forced emissions work with identical states and that the feature doesn't affect subsequent normal emissions

## Use Case
This feature enables scenarios where developers need to re-emit the current state to trigger listeners or observers, which is useful for state refresh operations or explicit state notifications.

### DIFF
--- a/packages/bloc/lib/src/bloc.dart
+++ b/packages/bloc/lib/src/bloc.dart
@@ -145,10 +145,15 @@ abstract class Bloc<Event, State> extends BlocBase<State>
   /// Updates the state of the bloc to the provided [state].
   /// A bloc's state should only be updated by `emitting` a new `state`
   /// from an [EventHandler] in response to an incoming event.
+  ///
+  /// Set [force] to `true` to emit the [state] even if it is equal to the
+  /// current [state].
   /// {@endtemplate}
   @visibleForTesting
   @override
-  void emit(State state) => super.emit(state);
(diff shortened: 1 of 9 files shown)


felangel:
This has been discussed at length in various issues and I have decided against it. Please avoid opening PRs without first searching the existing issues and/or opening a new issue. Thanks 🙏

